Issue #12, opened 2019-04-09
Implement MsBackendDataFrame


jorainer:
Implement the in-memory backend.

jorainer:
Basic class implemented. Remains to implement all required methods and add unit tests for them.

jorainer:
Rename that backend into `MsBackendDataFrame` (issue #9 )

lgatto:
@jorainer - have you renamed this in your own branch? If so, could you merge?

jorainer:
@lgatto , no, please wait for the pull request. I still have to finish implementing the last methods and add the related unit tests.

lgatto:
I have done some updates to MsBachend, mostly man, haven't touched anything else.

jorainer:
For the `MsBackendDataFrame` I would like the current implementation that does **not** initialize the `@spectraData` `DataFrame` with all supported columns but the accessors would take care of creating eventually missing data on-the-fly. An example accessor is shown below:

```r
setMethod("precScanNum", "MsBackendDataFrame", function(object) {
    if (any(colnames(object@spectraData) == "precScanNum"))
        object@spectraData$precScanNum
    else rep(NA_integer_, times = length(object))
})
```

My two main reasons why I think that this is a good implementation are:
1) we don't need to allocate memory for something that is not there. Keeps memory demand low.
2) the object is more flexible and is inherently backwards compatible (or forward compatible?): assume we require the `MsBackendDataFrame` to have all mandatory columns in its `@spectraData` `DataFrame` to be always initialized and present and a user saves a `MsBackendDataFrame` object to disk. When we add a field/variable to the list of mandatory columns in an updated version of the package, this column will be missing in the saved version and the user will get an error. Now, for the current implementation that creates missing column content on the fly that would be no problem at all. This would make maintanence much simpler, as we don't have to be carefull with class versions, updating objects etc.

Any thoughts on that @lgatto @sgibb ?

lgatto:
I am not convinced by 1, but you win me over with 2.

PR #16, opened 2019-04-12
Add implementations for MsBackendDataFrame and MsBackendMzR

jorainer:
This PR adds the implementations of the `MsBackendDataFrame` and the `MsBackendMzR` backends. Happy to discuss :)


R/MsBackendDataFrame.R
```diff
@@ -264,26 +281,24 @@ setReplaceMethod("spectraData", "MsBackendMemory", function(object, value) {
...
     object
 })
 
 #' @rdname hidden_aliases
-setMethod("spectraVariables", "MsBackendMemory", function(object) {
-    ## TODO @jo IMPLEMENT:
-    ## list the colnames of spectraData AND all columns supported by accessor
-    ## methods.
+setMethod("spectraVariables", "MsBackendDataFrame", function(object) {
+    unique(c(names(.SPECTRA_DATA_COLUMNS), colnames(object@spectraData)))
```
lgatto: Why do you take the union of the actual column names and the mandatory ones defined  in `.SPECTRA_DATA_COLUMNS`?

R/MsBackendDataFrame-functions.R
```diff
@@ -92,9 +101,53 @@ NULL
...
+#' retrieve the corresponding data.
+#'
+#' @param x object with a `@spectraData` slot containing a `DataFrame`.
+#'
+#' @param column `character(1)` with the column name.
+#'
+#' @author Johannes Rainer
+#'
+#' @noRd
+.get_spectra_data_column <- function(x, column) {
```
lgatto: I don't understand the rational for this one. In the `MsBackendDataFrame`, is all the data not supposed to be stored in `x@spectraData`? Why would there be two ways to access it, using `spectraData` and the accessor? Would the accessor not itself retrieve a `spectraData` column?

R/MsBackendDataFrame.R
```diff
@@ -95,80 +96,79 @@ setReplaceMethod("collisionEnergy", "MsBackendMemory", function(object, value) {
...
 })
 
 #' @rdname hidden_aliases
-setMethod("intensity", "MsBackendMemory", function(object) {
+setMethod("intensity", "MsBackendDataFrame", function(object) {
     if (any(colnames(object@spectraData) == "intensity"))
         object@spectraData$intensity
     else {
-        lst <- list(numeric)
+        lst <- list(numeric())
```
lgatto: Why not 
```r
rep(list(numeric()), times = length(object))
```

R/MsBackendDataFrame.R
```diff
@@ -95,80 +96,79 @@ setReplaceMethod("collisionEnergy", "MsBackendMemory", function(object, value) {
...
 })
 
 #' @rdname hidden_aliases
-setMethod("mz", "MsBackendMemory", function(object) {
+setMethod("mz", "MsBackendDataFrame", function(object) {
     if (any(colnames(object@spectraData) == "mz"))
         object@spectraData$mz
     else {
-        lst <- list(numeric)
+        lst <- list(numeric())
```
lgatto: Same as above for `intensity`.

R/MsBackendDataFrame.R
```diff
@@ -95,80 +96,79 @@ setReplaceMethod("collisionEnergy", "MsBackendMemory", function(object, value) {
...
         lst[rep(1, times = length(object))]
     }
 })
 
 #' @rdname hidden_aliases
-setMethod("peaks", "MsBackendMemory", function(object) {
+setMethod("peaks", "MsBackendDataFrame", function(object) {
     mapply(mz(object), intensity(object), FUN = function(m, i)
-        data.frame(mz = mz, intensity = i))
+        data.frame(mz = m, intensity = i), SIMPLIFY = FALSE, USE.NAMES = FALSE)
```
lgatto: I would suggest that we return a `matrix` rather than a `data.frame` here. Any reason why you favour the latter?

R/MsBackendDataFrame.R
```diff
@@ -248,14 +248,31 @@ setReplaceMethod("smoothed", "MsBackendMemory", function(object, value) {
...
 
 #' @rdname hidden_aliases
-setMethod("spectraData", "MsBackendMemory", function(object, columns) {
-    ## TODO @jo IMPLEMENT:
-    ## allow to specify columns. Have then to check if these columns exist,
-    ## if not call the respective method.
+#'
+#' @importFrom methods as
+setMethod("spectraData", "MsBackendDataFrame", function(object, columns) {
+    if (missing(columns))
```
lgatto: Same as above, I don't see why you take the union of actual and mandatory column names.

R/MsBackendDataFrame.R
```diff
@@ -248,14 +248,31 @@ setReplaceMethod("smoothed", "MsBackendMemory", function(object, value) {
 })
 
 #' @rdname hidden_aliases
-setMethod("spectraData", "MsBackendMemory", function(object, columns) {
-    ## TODO @jo IMPLEMENT:
-    ## allow to specify columns. Have then to check if these columns exist,
-    ## if not call the respective method.
+#'
+#' @importFrom methods as
+setMethod("spectraData", "MsBackendDataFrame", function(object, columns) {
```
lgatto: I anticipated to use `columns = TRUE` as default (see [here](https://github.com/lgatto/Spectra/blob/master/R/MsBackend.R#L495)), which returns all the columns. It is generally easier to have an explicit default rather than play with `missing` in the code. But my default clearly didn't anticipate the union of actual and mandatory columns.

R/MsBackendDataFrame.R
```diff
@@ -248,14 +248,31 @@ setReplaceMethod("smoothed", "MsBackendMemory", function(object, value) {
...
+            res$mz <- if (length(other_res$mz)) other_res$mz else list()
+        if (any(names(other_res) == "intensity"))
+            res$intensity <- if (length(other_res$intensity)) other_res$intensity else list()
+    }
+    res[, columns, drop = FALSE]
 })
 
 #' @rdname hidden_aliases
-setReplaceMethod("spectraData", "MsBackendMemory", function(object, value) {
+setReplaceMethod("spectraData", "MsBackendDataFrame", function(object, value) {
```
lgatto: Don't we need to check for column names here, making sure the mandatory ones are present?

jorainer:
Regarding the set of mandatory and optional columns: the mandatory columns are the columns for which we have also the accessor methods (they are also the slots we had in the `Spectrum` objects). Actually, they are not mandatory, we are just enforcing the data type on them. So, `msLevel` always has to be an `integer` etc.

I also want to allow optional columns because we need a way to add arbitrary annotations to each spectrum. So, the user is free to add whatever additional column he wants.

R/MsBackendDataFrame.R
jorainer: Each `MsBackend` **has** to provide the data from the mandatory columns (*via* accessor methods). In addition, the backend has to tell the user what other fields/variables it provides. Imagine you have a `MsBackendMassBank` that uses a SQL backend to retrieve spectra from a database. You want to know what variables/annotations you would get for a spectrum from the database. `spectraVaribles` should then return you the names of the mandatory variables and the ones it provides in addition, hence the union.

R/MsBackendDataFrame-functions.R
jorainer: I don't think it makes sense to keep all data (all mandatory columns) in memory in the `@spectraData` slot. I would keep that flexible. Also have a look at the accessor methods. They will create the expected data and initialize it with `NA` if the column is not in the `@spectraData` slot. Why should we keep a `DataFrame` full of `NA` values?

R/MsBackendDataFrame.R
lgatto: I think there's some misunderstanding here. You talk about `MsBackend`, I refer to `MsBackendDataFrame`. `.SPECTRA_DATA_COLUMNS` is defined in the later, hence my questions. Are we talking about different things here?

lgatto: We can clarify this on the hangout, if easier.

R/MsBackendDataFrame.R
jorainer: Good point. Thanks. Thought that one would be slower because it creates a new object each time - but was wrong. I'll change.

R/MsBackendDataFrame.R
jorainer: Also here.

R/MsBackendDataFrame.R
jorainer: Yes. Acessing a column in a `data.frame` is faster than of a `matrix`. Also, creation *should* be faster and if I am not completely wrong `datafrane$mz` does not copy the content/vector while `matrix[, 1]` does... I believe.

R/MsBackendDataFrame.R
jorainer: I was more thinking of allowing the user to specify which columns he wants to get returned. With database backends in mind.

lgatto: Sure, that was my understanding too: default returns all, setting `columns` to a set of valid names will only return those.

lgatto:
> I also want to allow optional columns because we need a way to add arbitrary annotations to each spectrum. So, the user is free to add whatever additional column he wants.

Of course.

R/MsBackendDataFrame.R
jorainer: That's what I don't want to have - I would like to keep the `@spectraData` flexible and that not all *mandatory* columns should be present. I would also not call them mandatory - it's rather columns that have a mandatory data type and columns/variables that each backend needs to provide - whether or not the backend stores them is up to the backend. My `MsBackendDataFrame` does not store columns that are purely `NA`.

lgatto: Ok, I see. So rather than having these columns filled with `NA`, you prefer to generate them on the fly every time these data need to be accessed. 

An option would be to create them if absent, and simplify all the accessors to return these columns. I need to think about this a bit more. Or we can discuss this.

R/MsBackendDataFrame.R
jorainer: But then I find it confusing to have the default `columns = TRUE`. If I was a user and see the definitions `columns = TRUE` I would assume that this parameter takes only a `logical(1)` as input, and not either `TRUE` or a `character` of column names ... but that's just a minor thing anyway.

R/MsBackendDataFrame.R
jorainer: Honestly, also looking at the size objects can get (I have a `OnDiskMSnExp` of *only* 1200 files where just the `fData` is 1.4GB large) I'd prefer the way to not require to always have all of these columns in the `@spectraData`.

lgatto: And could this be encoded as an `Rle`?
```r
> x <- rep(NA, 1e6)
> pryr::object_size(x)
4 MB
> pryr::object_size(Rle(x))
1.18 kB
```
I am thinking in terms of code (simpler accessors) and overhead of generating these long vectors repeatedly.

lgatto: Interestingly:
```r
> x <- rep(NA, 1e6)
> df1 <- DataFrame(x = x, y = x, z = x)
> df2 <- DataFrame(x = Rle(x), y = Rle(x), z = Rle(x))
> df3 <- DataFrame(x = rep(NA, 1e6), y = rep(NA, 1e6), z = rep(NA, 1e6))
> object_size(x)
4 MB
> object_size(Rle(x))
1.18 kB
> object_size(df1)
4 MB
> object_size(df2)
3.73 kB
> object_size(df3)
12 MB
```
Even without the `Rle`, if we use the same vector of `NA` values to the different variables, we don't multiply memory usage. This could even be reduced by reusing the Rle
```r
> xrle <- Rle(x)
> df4 <- DataFrame(x = xrle, y = xrle, z = xrle)
> object_size(df4)
2.74 kB
```

jorainer: I have to think over that. But also with the `Rle` there will be some more code necessary. We would have to un-rle if the column is an Rle in the accessors (which should be less of a problem) and also in the `spectraData` accessor. 

And I did not yet find a possibility to check the data type of an `Rle` - do you know how that could be done? `is.integer(Rle(NA_integer, 5))` is not working.

lgatto:
```r
> x <- Rle(rep(NA_integer_, 1e3))
> str(as.integer(x))
 int [1:1000] NA NA NA NA NA NA NA NA NA NA ...
> str(as.vector(x))
 int [1:1000] NA NA NA NA NA NA NA NA NA NA ...
```

lgatto: What we need to clarify here isn't really Rle or not, but whether it is easier (code and maintenance wise) and more/less efficient (repeated only the fly generation vs somewhat larger data) to have the initialise the `DataFrame`. Let's sleep over it and touch back base next week.

jorainer: In the meantime I started to like the `Rle` idea. For some columns like `fromFile` it could make sense because it reduces the size. The accessors would *only* have to call `as.vector(Rle(1L, 1e6))`. To check for the data type we could use `class(x@value)` where `x` is a `Rle`. So, performance wise it should be *doable*.

I might try to implement a `MsBackendRleDataFrame` and we can compare.

lgatto: Ok, but I don't think we will need `MsBackendDataFrame` and `MsBackendRleDataFrame` - I assume this is only for testing?

R/MsBackendDataFrame-functions.R
```diff
@@ -74,6 +74,15 @@ NULL
...
+        "Length of mz and intensity values differ for some spectra"
+    else NULL
+}
+
+#' data types of spectraData columns
+#'
+#' @noRd
 .SPECTRA_DATA_COLUMNS <- c(
     msLevel = "integer",
     rt = "numeric",
```
lgatto: I would suggest time call this `rtime` rather than `rt`, as `rt` is a function to generate a random sample from a t distribution.

R/MsBackendDataFrame-functions.R
```diff
@@ -92,9 +101,53 @@ NULL
     collisionEnergy = "numeric"
 )
 
+#' accessor methods for spectraData columns.
+#'
+#' @noRd
+.SPECTRA_DATA_COLUMN_METHODS <- c(
+    msLevel = "msLevel",
+    rt = "rtime",
```
lgatto: See above for rather choosing `rtime`.

R/MsBackendDataFrame.R
jorainer: Yes. I want to test how the `Rle`s work in the `spectraData` - especially for subsetting etc.

R/MsBackendDataFrame-functions.R
jorainer: agree

R/MsBackendDataFrame-functions.R
jorainer: OK

R/MsBackendDataFrame.R
lgatto: From the small test below, it looks like accessing a matrix column is faster, except if one uses `[[`.

```r
> data(itraqdata)
> dfr <- as(itraqdata[[1]], "data.frame")
> mat <- as.matrix(dfr)
microbenchmark(dfr[, 1], mat[, 1], dfr[, "mz"], mat[, "mz"], dfr[[1]], dfr[["mz"]])
Unit: microseconds
        expr   min      lq     mean  median      uq    max neval
    dfr[, 1] 9.779 10.3585 11.29819 10.8885 11.5125 45.016   100
    mat[, 1] 6.868  7.6930  8.23260  7.9920  8.6070 20.905   100
 dfr[, "mz"] 9.762 10.5505 11.18423 10.9155 11.7100 16.027   100
 mat[, "mz"] 6.803  7.6605  8.36979  8.0745  8.6530 33.461   100
    dfr[[1]] 4.964  5.6965  6.20307  6.0955  6.5190 10.753   100
 dfr[["mz"]] 5.038  5.7225  6.63447  6.1035  6.6800 39.812   100
```

Accessing a set of rows is considerably faster in matrices:

```r
> microbenchmark(dfr[20:99, ], mat[20:99, ], dfr[20:99, 1], mat[20:99, 1], dfr[[1]][20:99])
Unit: microseconds
            expr    min      lq     mean  median      uq     max neval
    dfr[20:99, ] 44.337 45.6870 47.01528 46.3155 46.9885  85.723   100
    mat[20:99, ]  2.069  2.5240  2.90010  2.8465  3.1115  10.600   100
   dfr[20:99, 1] 18.606 19.6800 21.57392 20.3080 20.9350 125.570   100
   mat[20:99, 1]  1.700  2.2745  2.61878  2.6245  2.9510   3.799   100
 dfr[[1]][20:99]  7.022  7.7185  8.64923  8.6995  9.0185  21.733   100
```
My initial suggestion to use a matrix rather than a data.frame was that we had only numerics. The above seems to suggest matrices would be better too. There might be other reasons to use dataframes.

What do you think?

jorainer: OK, then we go for `matrix`.

R/MsBackendDataFrame.R
lgatto: Here's a explanation is issues with `missing()`. Instead of `TRUE`, we could use `NULL` (as in the video), or `spectraVariables(object)`, to make it explicit that it's a character with spectra variables.

jorainer: 👍 `columns = spectraVariables(object)`

R/MsBackendDataFrame.R
jorainer: The performance comparison is in issue #17 
Regarding implementation notes we should discuss in issue #12

lgatto: As said in #12, I am happy with on-the-fly construction of data in the accessors.

Issue #11, opened 2019-04-09
Implement MsBackendMzR


jorainer:
Implement the `MsBackendMzR` class.

jorainer:
Implementation at present commented out, but general structure there.

jorainer:
`MsBackendMzR` extends `MzBackendDataFrame`. Has been added in PR #16.

jorainer:
Would be nice to pick the idea of the `Rle` up here. From `mzR` we get a lot of header columns most of which are however `NA` in many cases. Would be nice to reduce the size of the `@spectraData` `DataFrame` with @lgatto 's suggestion to use `Rle`s here. Eventually we might still want to have the `MsBackendRleDataFrame` and `MsBackendMzR` extending that backend instead of the `MsBackendDataFrame`.

jorainer:
Move the `Rle` functionality into the `MsBackendMzR` object and drop the `MsBackendRleDataFrame`.

PR #19, opened 2019-04-23
Add MsBackendDataFrame and MsBackendMzR

jorainer:
Add two backends:
- `MsBackendDataFrame` (issue #12): keeps all the data in memory. Allows missing *mandatory* columns in the `DataFrame` creating their content on-the-fly in the accessor methods (see also issue #12).
- `MsBackendMzR` (issue #11): extends `MsBackendDataFrame` reads the content of `mzR::header` into its `@spectraData` `DataFrame` while reading m/z and intensity values from the original files on demand. Uses `Rle` to *compress* data content (columns with only single values or `NA`) to reduce memory demand (since most of the columns returned by `mzR::header` will be empty or `0` for many input files and formats). For missing mandatory columns this backend also creates their content on the fly.


R/MsBackend.R
```diff
@@ -27,36 +27,44 @@ NULL
...
 #'     total ion current should be (re)calculated on the actual data
 #'     (`initial = FALSE`).
 #'
+#' @param i For `[`: `integer`, `logical` or `character` to subset the object.
+#'
+#' @param j For `[`: not supported.
+#'
+#' @param object Object extending `MsBackend`.
+#'
+#' @param spectraData For : `DataFrame` with spectrum metadata/data.
```
lgatto: I think something's missing here.

R/MsBackend.R
```diff
@@ -27,36 +27,44 @@ NULL
...
+#'
 #' @param value replacement value for `<-` methods. See individual
 #'     method description or expected data type.
 #'
+#' @param x Object extending `MsBackend`.
```
lgatto: I assume that the signatures of the generics are different, but it's kind of annoying that there's `x` and `object` that refer to the same thing. And I guess it isn't possible to at least put them together using roxygen?

R/MsBackendMzR-functions.R
```diff
@@ -21,14 +21,118 @@ MsBackendMzR <- function() {
...
+#' @return `DataFrame`
+#'
+#' @author Johannes Rainer
+#'
+#' @importClassesFrom S4Vectors Rle
+#'
+#' @importFrom S4Vectors Rle
+#'
+#' @noRd
+.compress_spectra_data <- function(x, columns = c("fromFile")) {
```
lgatto: I am not convinced with calling this *compressing* a column. Wouldn't it be clearer to calls this `.as_rle_spectra_data`. Similarly replace `.uncompress_spectra_data` with `.as_vector_spectra_data`?

I've had the same thought [here](https://github.com/lgatto/Features/blob/0b85d90a47b6e4b2feb95df03b15af3ec7f3f2a0/R/shrink.R#L75) that I initially called compressing a column, but dropped it (not confined that *shrinking* is better though) because compression has a specific meaning.

R/MsBackend.R
jorainer: Of course. I'll add the missing part.

R/MsBackend.R
jorainer: Yes the signatures for the generics are different. I usually try to always use `object`, but `length` etc just have `x`. I am not aware of a way to combine them with roxygen.

R/MsBackendMzR-functions.R
jorainer: Agree. `.as_rle_` and `.as_vector_` exactly tells what's doing, so I'll use that.

jorainer: Or `.deflate_`, `.inflate` might also be nice - deflating a blown-up `DataFrame` with just *air*

jorainer:
You're right - terminal is working but is unreadable, emacs is dying. I have no issues with the `DataFrame` returned by `spectraData` to have columns `mz` and `intensity` as `SimpleList` - the question is what to do with the `mz` and `intensity` methods - should they return also a `SimpleList` instead of a `list`?

jorainer:
For consistency reasons it might be better to use and require always `SimpleList` for `mz` and `intensity` for any backends.

lgatto:
Yes, as we are using the `S4Vectors`/`DataFrame` infrastructure, I think we should be consistent with it, and use `SimpleList` throughput. 

It however comes with an overhead:
```r
> x <- rnorm(1e6)
> microbenchmark(list(x), SimpleList(x))
Unit: nanoseconds
          expr   min    lq     mean  median      uq    max neval
       list(x)    77    91   209.80   200.0   223.0   3885   100
 SimpleList(x) 91278 92902 98957.09 94060.5 98713.5 231358   100
> microbenchmark(as.list(x), SimpleList(as.list(x)))
Unit: milliseconds
                   expr      min       lq     mean   median       uq      max
             as.list(x) 69.25408 116.6951 122.2446 128.1850 131.0231 138.8164
 SimpleList(as.list(x)) 23.58498 129.2324 126.7507 134.2477 136.4811 166.7643
 neval
   100
   100
```
In addition to the nice printing, we also benefit from many sub-classes:
```r
> NumericList(as.list(x))
NumericList of length 1000000
[[1]] -0.805444824795359
[[2]] -1.13692475440835
[[3]] -0.625832055706986
[[4]] 0.660549439979912
[[5]] -0.590670393977926
[[6]] -1.09335668732407
[[7]] -1.90779307889393
[[8]] -0.240546671329855
[[9]] 0.591808756265667
[[10]] -0.176166086075084
...
<999990 more elements>
```
and `IRangesList`, `RleList`, and probably more that will become essential later on.

jorainer:
Where is the `NumericList` constructor defined?

lgatto:
Oh, it's in `IRanges`.

jorainer:
Uh, and performance-wise `NumericList` is, well, not fast:

```r
> microbenchmark(NumericList(x), SimpleList(x), list(x))
Unit: nanoseconds
           expr     min      lq       mean    median      uq      max neval cld
 NumericList(x) 4372855 5680430 7062882.41 6231177.5 6489534 17363068   100   b
  SimpleList(x)  113877  146947  158739.94  157623.5  173550   313134   100  a 
        list(x)      73     134     350.53     328.5     478     1031   100  a 

```

lgatto:
What about:
```r
> microbenchmark(NumericList(x), NumericList(x, compress = FALSE), SimpleList(x), list(x), as(SimpleList(x), "NumericList"))
Unit: nanoseconds
                             expr     min      lq       mean    median
                   NumericList(x) 2447088 2485352 2809701.40 2528291.5
 NumericList(x, compress = FALSE)  649264  661153  706043.12  669612.0
                    SimpleList(x)   93930  100516  105394.63  102644.5
                          list(x)      84     325     369.42     380.5
 as(SimpleList(x), "NumericList")  527252  541752  564833.99  551267.0
        uq     max neval
 2924080.5 4090867   100
  680067.0 2210639   100
  107443.5  197201   100
     440.5    1189   100
  565732.5 1090455   100
```
Still not fast, but not as terrible. 

The advantage of the sub-classes is that we get some guarantee that the elements in the list make sense.

I'm open to discuss all this.

jorainer:
I don't have a strong opinion on whether or not to use `NumericList`. I tend a little towards `SimpleList` because a) we don't need to import/depend also on `IRanges` and b) construction takes longer - and that's something we use/need a lot for the `MsBackendMzR` backend (and others as well).

lgatto:
On the other hand, there are only 2 columns, `mz` and `intensity`for with we need these - does it make much different for these two columns out of tens?

jorainer:
For the `MsBackendDataFrame` it will not matter that much, because it will be converted once, at initializiation. The `MsBackendMzR` needs to call `NumericList` for each call to `mz` and `intensity` as these values are read from the raw files.

Apart from the not required validation of the list elements' data type, is there any other benefit you know of?

lgatto:
> Apart from the not required validation of the list elements' data type, is there any other benefit you know of?

None I can think of. But, if the plan is to validate the data every time the accessor is called:
```r
> x <- as.list(sample(1e5))
> microbenchmark(all(sapply(x, is.integer)), NumericList(x))
Unit: milliseconds
                       expr       min       lq     mean   median       uq
 all(sapply(x, is.integer)) 19.854548 20.56740 22.28687 21.66394 22.75967
             NumericList(x)  9.580873  9.88326 10.77660 10.34341 10.96147
      max neval
 58.35280   100
 16.32579   100
```
If we skip validation when accessing mz and intensities, then I guess `SimpleList` is enough.

jorainer:
None of the accessors checks for correct data.

For the `MsBackendMzR` we rely on what `mzR` returns and don't check the m/z and intensity values. `MsBackendDataFrame`'s `validObject` checks that all values are `numeric`, that lengths are the same for m/z and intensity and that m/z are ordered.

I'd say for now we keep `SimpleList` and eventually replace later.

jorainer:
Note: I made the requested changes in commit https://github.com/lgatto/Spectra/pull/19/commits/b6356f91e3220601c5bc7d454d38d811b89efdef

lgatto:
@jorainer - there's currently a conflict, but I'm not sure who's right. In `master`, there's a `spectraData` parameter for the `DataFrame` backend. Is it needed?

jorainer:
if in doubt, it's always me who's right (just kidding ;)) - I'll fix the conflict

jorainer:
Now, the `@param spectraData` did not make much sense there in the subsection for the `MsBackendDataFrame`. I would rather keep all parameters in one place and describe in the section for each subset exactly with which parameters a backend can (or needs to be) created/initialized.

In the recent commit I added the correct documentation for the `@param spectraData`.

sgibb:
I was wondering why the `?NumericList` says:
> CompressedList object the list elements are concatenated
together in a single vector-like object. The partitioning of
this single vector-like object (i.e. the information about where
each original list element starts and ends) is also kept in the
CompressedList object. This internal representation is generally
more memory efficient than SimpleList, especially if the object
has many list elements (e.g. thousands or millions). Also it makes
it possible to implement many basic list operations very
efficiently.

I can't see any memory nor performance benefit:
```r
library("IRanges")
library("S4Vectors")
library("msdata")
library("mzR")
library("microbenchmark")

sciex <- dir(system.file("sciex", package = "msdata"), full.names = TRUE)[1]
fh <- openMSfile(sciex)
p <- peaks(fh)
close(fh)
mz <- lapply(p, "[", , 1L)

microbenchmark(
    createSL=SimpleList(mz),
    createNLC=NumericList(mz),
    createNLU=NumericList(mz, compress=FALSE)
)
# Unit: microseconds
#       expr       min        lq       mean     median        uq        max neval
#   createSL   131.996   150.977   215.6794   170.0075   191.710   2381.956   100
#  createNLC 10369.469 10728.606 17026.6877 13659.8350 15529.400 126174.549   100
#  createNLU  4740.780  4986.514  5922.2170  5143.1285  5742.184  13176.855   100

set.seed(2019)
i <- sample(length(mz), size=round(length(mz)/3))

sl <- SimpleList(mz)
nlc <- NumericList(mz)
nlu <- NumericList(mz, compress=FALSE)

lapply(list(mz, sl, nlc, nlu), object.size)
# [[1]]
# 13490616 bytes
#
# [[2]]
# 13491800 bytes
#
# [[3]]
# 13444920 bytes
#
# [[4]]
# 13491808 bytes
#

microbenchmark(
    indexBL=mz[i],
    indexSL=sl[i],
    indexNLC=nlc[i],
    indexNLU=nlu[i]
)
# Unit: microseconds
#      expr      min        lq       mean    median        uq       max neval
#   indexBL    2.302    3.4685    7.40452    4.1625    6.4770    36.670   100
#   indexSL  436.469  455.2020  622.24328  490.8520  562.3260  6232.492   100
#  indexNLC 3547.674 3656.2345 4396.06236 3784.5165 4322.1715 13290.953   100
#  indexNLU  433.205  449.4860  576.55831  488.9485  571.3275  3068.817   100

microbenchmark(
    replaceBL=mz[i] <- NA,
    replaceSL=sl[i] <- NA,
    replaceNLC=nlc[i] <- NA,
    replaceNLU=nlu[i] <- NA
)
# Unit: microseconds
#        expr       min         lq        mean     median         uq        max
#   replaceBL     2.693     6.5295     7.21697     7.2895     8.5385     29.372
#   replaceSL  3686.034  3851.8230  4690.36364  3991.3845  4287.0065  15268.382
#  replaceNLC 12229.823 14350.2805 24338.96566 15575.1910 20831.1630 212474.636
#  replaceNLU  3689.393  3847.4110  4425.49857  3976.7560  4209.7740  10848.160
#  neval
#    100
#    100
#    100
#    100

microbenchmark(
    lapplyBL=lapply(mz[i], "[", 1),
    lapplySL=lapply(sl[i], "[", 1),
    lapplyNLC=lapply(nlc[i], "[", 1),
    lapplyNLU=lapply(nlu[i], "[", 1)
)
# Unit: microseconds
#       expr      min        lq       mean    median        uq      max neval
#   lapplyBL   78.331   84.7850   91.84887   87.6160   93.8155  172.922   100
#   lapplySL  528.346  571.4165  641.20262  597.3685  638.4410 3203.489   100
#  lapplyNLC 1867.936 1971.5950 2176.27551 2053.0220 2155.2425 8748.972   100
#  lapplyNLU  530.697  567.6730  652.63258  600.7910  650.0940 2864.165   100
```

jorainer:
That's why I'd say we use `compress = FALSE`.

Issue #20, opened 2019-04-26
Spectra interface


jorainer:
The `Spectra` interface should provide data accessor, manipulation and filtering/subsetting methods.

## Data accessors

Most of these will simply call the corresponding method from the `MsBackend`.

- [ ] `$`: access a column in `@spectraData`.
- [ ] `acquisitionNum`
- [ ] `bpi`
- [ ] `centroided`, `centroided<-`
- [ ] `chromatogram` (eventually later)
- [ ] `collisionEnergy`
- [ ] `fileNames`: content of `@files`
- [ ] `fromFile`
- [ ] `intensity`: here we need to execute the processing queue
- [ ] `ionCount`
- [ ] `isCentroided`
- [ ] `isEmpty`
- [ ] `length`
- [ ] `metadata`
- [ ] `msLevel`
- [ ] `mz`
- [ ] `polarity`
- [ ] `peaksCount`
- [ ] `precursorCharge`
- [ ] `rtime`, `rtime<-`
- [ ] `scanIndex`
- [ ] `smoothed`,`smoothed<-`
- [ ] `spectraData`
- [ ] `spectraNames`
- [ ] `spectraVariables` (colnames of `spectraData`)
- [ ] `tic`

## Data manipulators

- [ ] `bin`
- [ ] `clean`
- [ ] `compareSpectra`
- [ ] `combineSpectra`
- [ ] `estimateMzResolution`
- [ ] `estimateNoise`
- [ ] `normalize`
- [ ] `pickPeaks`
- [ ] `quantify`
- [ ] `removeReporters`
- [ ] `removePeaks`
- [ ] `smooth`

## Data filtering and subsetting

- [ ] `[`
- [ ] `[[` - what should that return? a `Spectrum`?
- [ ] `filter`: general purpose filter function that could use `AnnotationFilter` a-la: `filter(sps, ~ polarity == 1 & rtime >= 12.2 & rtime <= 12.4)`
- [ ] `filterAcquisitionNum`
- [ ] `filterEmptySpectra`
- [ ] `filterFile`
- [ ] `filterMsLevels`
- [ ] `filterMz`
- [ ] `filterPrecursorScan`
- [ ] `filterPolarity`
- [ ] `filterRt`

sgibb:
I am going to implement `Spectra,normalize`.

jorainer:
Can you eventually also implement the `pickPeaks`? I guess there will be some code that can go to `MsCoreUtils` - and you know it best.

PR #22, opened 2019-05-13
First working version of Spectra

jorainer:
This PR contains the first working `Spectra` implementation and defines many the `Spectra` interface (issue #20).

Most of the methods are simple getter/setter methods.

What is worth checking is the processing queue. It's the same concept as we have it in `OnDiskMSnExp`: all data manipulation methods add themselfs to the `processingQueue` which gets executed each time a method accesses `mz` or `intensity` values. What's new is that we can no longer use the spectra functions (like `clean_Spectrum`) because we don't have `Spectrum` objects anymore. The new functions to apply the changes are defined in *Spectra-functions.R*. Currently there are *only* `.clean_peaks` and `.remove_peaks`. Such functions should take the following arguments:
- `x`: `matrix` with mz and intensity values.
- `spectrumMsLevel`: the MS level of the spectrum - to allow the manipulation to be applied to spectra of a certain MS level only.
- `centroided`: optional; some functions might need it.
- `msLevel.`: to which MS level should the function be applied.

To add arbitrary spectra annotation columns one can use: `spectraData(x)$new_column <- `. We might eventually simplify this later (e.g. with `x$new_column <-`).


jorainer:
@sgibb I added you as a reviewer, but please only review if you have enough time.

jorainer:
Closing PR - explain and discuss implementations first in conference call.

PR #33, opened 2019-06-12
Use Rle in .as_rle whenever possible and useful

sgibb:
As disscussed in:
https://github.com/rformassspectrometry/Spectra/pull/31/files#r291991241

There are some failing unit tests:
- test_MsBackendDataFrame.R:61 before merging backends the data in the columns arn't converted to `Rle` because the `Rle` would be larger/of equal size than the bare vector. After merging we could save some space by using `Rle`s. Maybe it would be useful to call `.as_rle` after merging?
https://github.com/rformassspectrometry/Spectra/blob/9e35a0b11320bf9e671474079981ce33d7af15ed/tests/testthat/test_MsBackendDataFrame.R#L61-L62

- HDF5 test because of missing `h5files` slot.


jorainer:
I'd say we merge and evaluate then whether we call `.as_rle` after merging or not.

PR #38, opened 2019-06-18
Require rt to be of length 2 in filterRt

lgatto:
I have added a check for rt to be `numeric(2)` and `rt[2] > rt[1]`.


DESCRIPTION
```diff
@@ -33,7 +33,8 @@ Suggests:
     BiocStyle (>= 2.5.19),
     mzR (>= 2.15.1),
     rhdf5,
-    rmarkdown
+    rmarkdown,
+    magrittr
```
jorainer: Do we need `magrittr`? I guess for the vignette?

R/Spectra.R
```diff
@@ -1028,6 +1028,8 @@ setMethod("filterPrecursorScan", "Spectra",
 #' @rdname Spectra
 setMethod("filterRt", "Spectra",
           function(object, rt = numeric(), msLevel. = unique(msLevel(object))) {
+              if (!identical(length(rt), 2L) || rt[1] >= rt[2])
+                  stop("Please provide a lower and upper retention time range.")
               suppressWarnings(rt <- range(rt))
```
jorainer: Just a comment: I used here the `range(rt)` to silently ensure that length rt is always 2 and that `rt[2] > rt[1]`. If you prefer to have the explicit check here we can remove that line.

lgatto: This PR came up because I initially tried it with a single value, and I hadn't been warned of my mistake, which would be beneficial, I think.

jorainer: I'm OK with that.

R/Spectra.R
```diff
@@ -1028,6 +1028,8 @@ setMethod("filterPrecursorScan", "Spectra",
 #' @rdname Spectra
 setMethod("filterRt", "Spectra",
           function(object, rt = numeric(), msLevel. = unique(msLevel(object))) {
+              if (!identical(length(rt), 2L) || rt[1] >= rt[2])
```
sgibb: I know this is a small function that won't be called often but isn't `length(rt) != 2L` enough? Also this should check for `is.numeric`. Otherwise the user could use `characters` (won't raise an error) or `Date`s.
```suggestion
              if (length(rt) != 2L || !is.numeric(rt) || rt[1] >= rt[2])
```

R/Spectra.R
sgibb: If we ensure that `rt` is numeric and in the correct order by the `if` statement above do we really need `suppressWarning` here. Also `range` should be superfluous. IMHO we can remove the whole line.
```suggestion
```

tests/testthat/test_Spectra.R
```diff
@@ -861,14 +861,16 @@ test_that("filterPrecursorScan,Spectra works", {
...
+    res <- filterRt(sps, c(1, 2))
     expect_true(is(res, "Spectra"))
     expect_true(length(res) == 0)
     expect_true(length(res@processing) == 1)
 
     sps <- Spectra(sciex_mzr)
     res <- filterRt(sps, rt = c(100, 120))
     expect_true(all(rtime(res) >= 100 & rtime(res) <= 120))
+    expect_error(filterRt(sps, rt = c(100)))
+    expect_error(filterRt(sps, rt = c(120, 100)))
```
sgibb: We should add a test with non-numerics as well.

R/Spectra.R
lgatto: You are right. I wrongly assumed `rt` was defined as a numeric in the signature. But
```r
> filterRt(sps, c(1, 2))
MSn data (Spectra) with 0 spectra in a MsBackendDataFrame backend:
Processing:
 Filter: select retention time [1..2] on MS level(s)  [Tue Jun 18 20:46:26 2019] 
> filterRt(sps, c("1", "2"))
Error in .local(object, ...) : 'rt' must be a numeric
> traceback()
7: stop("'rt' must be a numeric")
6: .local(object, ...)
5: filterRt(object@backend, rt, msLevel.)
4: filterRt(object@backend, rt, msLevel.) at Spectra.R!7751MA3#6
3: .local(object, ...)
2: filterRt(sps, c("1", "2"))
1: filterRt(sps, c("1", "2"))
> getMethod("filterRt", "MsBackendDataFrame")
Method Definition:

function (object, ...) 
{
    .local <- function (object, rt = numeric(), msLevel = integer()) 
    {
        if (length(rt)) {
            rt <- range(rt)
            if (!is.numeric(rt)) 
                stop("'rt' must be a numeric")
            if (!length(msLevel)) 
                msLevel <- unique(msLevel(object))
            sel_ms <- msLevel(object) %in% msLevel
            sel_rt <- rtime(object) >= rt[1] & rtime(object) <= 
                rt[2] & sel_ms
            object[sel_rt | !sel_ms]
        }
        else object
    }
    .local(object, ...)
}
<bytecode: 0x55aa2b8cc168>
<environment: namespace:Spectra>

Signatures:
        object              
target  "MsBackendDataFrame"
defined "MsBackendDataFrame"
```
This is tested in the `MsBackendDataFrame` method. It should probably be defined once in `filterRt,Spectra`, shouldn't it?

This raises the question whether we would want to do these checks more systematically.

jorainer: Well, yes. Would make sense to do all input parameter tests in `Spectra`. As of now they are in the backends because I first implemented the backends and then `Spectra`, but we could move all tests up.

lgatto: Yes, let's aim for that. It makes much more sense, as the API is defined at the `Spectra` level anyway.

lgatto:
I have done the additional suggestions. But there's something else.
In 
```r
#' @rdname hidden_aliases
setMethod("filterRt", "MsBackendDataFrame",
          function(object, rt = numeric(), msLevel = integer()) {
              if (length(rt)) {
                  rt <- range(rt)
                  if (!length(msLevel))
                      msLevel <- unique(msLevel(object))
                  sel_ms <- msLevel(object) %in% msLevel
                  sel_rt <- rtime(object) >= rt[1] &
                      rtime(object) <= rt[2] & sel_ms
                  object[sel_rt | !sel_ms]
              } else object
          })
```
and 
```r
setMethod("filterRt", "Spectra",
          function(object, rt = numeric(), msLevel. = unique(msLevel(object))) {
              if (!is.numeric(msLevel.))
                  stop("Please provide a numeric MS level.")
              if (length(rt) != 2L || !is.numeric(rt) || rt[1] >= rt[2])
                  stop("Please provide a lower and upper numeric retention time range.")
              suppressWarnings(rt <- range(rt))
              object@backend <- filterRt(object@backend, rt, msLevel.)
              object@processing <- .logging(
                  object@processing,
                  "Filter: select retention time [", rt[1], "..", rt[2],
                  "] on MS level(s) ", paste0(msLevel., collapse = " "))
              object
          })
```
the defaults the `msLevel.` are different. Shouldn't we harmonise this?

jorainer:
We *could* - I wouldn't say we necessarily should. The normal user should never use methods from the backends, these are only for package developers. But it wouldn't hurt if we did, I would just not put that very high on the agenda.

lgatto:
> We _could_ - I wouldn't say we necessarily should. The normal user should never use methods from the backends, these are only for package developers. But it wouldn't hurt if we did, I would just not put that very high on the agenda.

I know users won't need to use backend methods, but from a consistency and maintenance point of view, default should be the same for every backend (unless there is a good reason not to) and this handled in Spectra. If there's an exception, it's handled in the backend method.

jorainer:
Agreed. Are you going to change that? Otherwise I think you're fine to merge.

lgatto:
Wait for a last addition that sets the same default and fixes a bug. Should be coming in a couple of minutes.

PR #40, opened 2019-06-20
Add applyProcessing function

jorainer:
Add `applyProcessing` function for `Spectra` to allow applying the lazy evaluation queue and writing modified peak data back to the data storage (if the backend supports that).


R/Spectra-functions.R
```diff
@@ -176,3 +176,36 @@ addProcessing <- function(object, FUN, ...) {
...
+                                            centroided(z), queue)
+        z
+    }, queue = object@processingQueue, BPPARAM = BPPARAM)
+    bknds <- backendMerge(bknds)
+    if (is.unsorted(f))
+        bknds <- bknds[order(unlist(split(seq_along(bknds), f),
+                                    use.names = FALSE))]
+    object@backend <- bknds
+    object@processing <- .logging(object@processing,
+                                  "Apply processing queue with ",
```
lgatto: Should this not say `"Applied processing queue with ..."`?

R/Spectra-functions.R
```diff
@@ -176,3 +176,36 @@ addProcessing <- function(object, FUN, ...) {
...
+        stop("length 'f' has to be equal to the length of 'object' (",
+             length(object), ")")
+    bknds <- bplapply(split(object@backend, f = f), function(z, queue) {
+        peaks(z) <- .apply_processing_queue(peaks(z), msLevel(z),
+                                            centroided(z), queue)
+        z
+    }, queue = object@processingQueue, BPPARAM = BPPARAM)
+    bknds <- backendMerge(bknds)
+    if (is.unsorted(f))
+        bknds <- bknds[order(unlist(split(seq_along(bknds), f),
```
sgibb: Just a quick question: Could we put this into `MsBackend::backendMerge` to avoid writing this construct multiple times in different places?

R/Spectra-functions.R
jorainer: Yep. I will change that.

R/Spectra-functions.R
jorainer: Which part would you like to put in the backend? The `order(...)`? I'm a little hesitant, because the backends should not know anything about being splitted and how. In addition I would want to keep this in `Spectra` because then implementation is in our hands - otherwise this would have to be done in each `backendMerge` of any backend - assuming the developer actually does take care of re-ordering them properly.

sgibb: You are right that we can't assume that any backend developer will take care of this. From the design point of view the ordering is part of the backend (no backend operation should change the order). That's why I suggest to put it into `MsBackend::backendMerge` than every backend developer has just to call `callNextMethod`. But this will change `MsBackend` from an interface to a class (and will not force the developer of backends to implement `backendMerge`).

Currently we use this `order` construct just in `applyProcessing` and `setBackend`. So I think we could keep it that way.

R/Spectra-functions.R
jorainer: Fixed

R/Spectra-functions.R
jorainer: Totally agree with you @sgibb . Would be clean to have that in the backend, but on the other side, the `Spectra` does something here the `MsBackend` doesn't know about: it is splitting the `MsBackend` by file to process them in parallel. I would want to have this parallel processing functionality in `Spectra`, so we have control over it in a central place.

Eventually we might find someday a better solution.

PR #45, opened 2019-07-15
Add matchApprox utility function

jorainer:
Add functionality helpful for spectra comparison:
  - Add matchApprox utility function that performs approximate numeric matching.
  - Add internal utility function .collocate_ordered_vectors.

These functions allow to match peaks from two spectra without the need to bin spectra beforehand: peaks can be matched between two spectra accepting a certain constant tolerance, and/or a user defined `ppm`. A correlation of intensity values between two spectra could thus be performed after first matching peaks' m/z values between spectra.

The core of the `matchApprox` function is implemented in C++ and is as fast as `match` (if not slightly faster).


jorainer:
Awg! well, I was not aware of that function, thanks for telling me @sgibb

sgibb:
> Was there any special reason that this function was never used in `compareSpectra`? In `MSnbase` `compareSpectra` enforces binning ...

To be honest: I can't remember. I wrote both (`relaxedMatch` for fragment matching in `MSnbase` and `synapter`; `compareSpectra`) at the same time. I guess it was because these days I was still influenced by my MALDI background where we had bad resolution and the vendor software did the binning anyway. 
If you use the default argument `method="common"` the common peaks are calculated by `match.closest` and binning is neither required nor done. Binning is just needed for `cor` and `dotproduct`. Implementing both without binning needs some thoughts (for peaks that could not matched).

jorainer:
Closing as we'll use the functionality from `MsCoreUtils` instead.

Issue #46, opened 2019-07-16
compareSpectra,Spectra,Spectra


jorainer:
I had some thoughts on the `compareSpectra` methdod (that also led to PR #45): I would suggest the `compareSpectra` to use peak matching with e.g. `match.closest` by default before applying the method to compare the spectra (on only the matching peaks). This would work also if the data was binned before. But, at least for the application in metabolomics, binning is not an option. People rather use a certain `ppm` to match peaks and compare spectra.

Any thoughts @lgatto and @sgibb ?

jorainer:
A very nice example for the use of binning in `compareSpectra` from @sgibb:

```r
x <- cbind(1:10, 1:10)
y <- cbind(9:20, 9:20)

Spectra:::.peaks_compare_intensities(x, y)
# [1] 1

xb <- bin(x[, 2], x[, 1], size=1, breaks=1:20)
yb <- bin(y[, 2], y[, 1], size=1, breaks=1:20)
xb <- cbind(xb$mids, xb$x)
yb <- cbind(yb$mids, yb$x)

Spectra:::.peaks_compare_intensities(xb, yb)
# [1] -0.552357
```

So, without binning the correlation is perfect because only the two overlapping peaks are compared. With binning the correlation is totally different. The binning assigns a value of `0` to each peak that is not present in one of the two. My gut feeling tells me that assuming an intensity of `0` for a peak that was not there (because absent, or because the m/z was simply outside the measurement range) is not OK, but I might be wrong here.

I think this is a very important issue that we need to solve. 
Would also be nice to get some feedback from e.g. @sneumann @michaelwitting @meowcat

sgibb:
An idea might be penalise the comparison results somehow, e.g. by the number of peaks used for it. In the example above it would be `cor * common/all_peaks` = `1 * 2/20`.

jorainer:
That would be one option - only, the result would then no longer be e.g. a valid correlation coefficient. I wonder if there is no literature on that?

meowcat:
Hm; I am not completely sure what I am being asked :)

For small molecule HRMS we typically don't want to bin, but align the spectra using a ppm window for matching, as noted above. The typical match that is used in NIST MS (originally for GC, but it propagated to the LC world) and therefore everyone copies it, is the modified cosine. 

This was originally written for unit mass resolution and therefore a priori "binned" spectra. As I understand, they work the same for HRMS spectra. NIST shows forward and reverse match.

[...]

PR #48, opened 2019-07-19
Add compareSpectra and re-use binning from MsCoreUtils

jorainer:
- Use `MsCoreUtils::bin` for binning.
- Add `compareSpectra` (issue #46) that by default uses the `MsCoreUtils::closest` for peak matching.

`compareSpectra,Spectra,Spectra` performs pairwise comparisons of all spectra in `x` with spectra in `y`. The result is a `matrix`, with `nrow(res) == length(x)` and `ncol(res) == length(y)`.

Some explanations about the parameters: I decided to have two parameters defining the tolerance in matching peaks: `tolerance`, which allows to specify a constant factor and `ppm` that allows to define a peak (m/z)-relative max allowed difference.


DESCRIPTION
```diff
@@ -42,6 +42,7 @@ VignetteBuilder: knitr
 BugReports: https://github.com/RforMassSpectrometry/Spectra/issues
 URL: https://github.com/RforMassSpectrometry/Spectra
 biocViews: Infrastructure, Proteomics, MassSpectrometry, Metabolomics
+Remotes: RforMassSpectrometry/MsCoreUtils
```
sgibb: Is that needed?

DESCRIPTION
```diff
@@ -55,6 +56,7 @@ Collate:
     'MsBackendMzR-functions.R'
     'MsBackendMzR.R'
     'ProcessingStep.R'
+    'RcppExports.R'
```
sgibb: Where does this come from? Left over?

R/RcppExports.R
```diff
@@ -0,0 +1,7 @@
+# Generated by using Rcpp::compileAttributes() -> do not edit by hand
+# Generator token: 10BE3573-1514-4C36-9D1C-5A225CD40393
+
+match_approx_cpp <- function(x, table, nomatch, tolerance) {
```
sgibb: Left over, I guess.

R/AllGenerics.R
```diff
@@ -12,10 +12,15 @@ setGeneric("backendMerge", def = function(object, ...)
     standardGeneric("backendMerge"),
     valueClass = "MsBackend")
 #' @rdname hidden_aliases
-setGeneric("bin", function(object, ...) standardGeneric("bin"))
+setGeneric("bin", function(x, ...) standardGeneric("bin"))
+#' @rdname hidden_aliases
+setMethod("bin", "numeric", MsCoreUtils::bin)
```
sgibb: Just thinking out loud here. Could it be that the following will be faster?
```r
.bin <- MsCoreUtils::bin
setMethod("bin", "numeric", .bin)
```

If you remember our benchmarks on slack:
```r
x <- 1:1e4
f1 <- function(x) { for(i in x)base::sort(x) }
f2 <- function(x) { for(i in x)sort(x) }
f3 <- function(x) { .sort <- base::sort; for(i in x).sort(x) }
microbenchmark(always=f1(x), never=f2(x), once=f3(x))
# Unit: milliseconds
#    expr   min    lq  mean median     uq    max neval
#  always 77.69 81.93 91.54  84.12 101.91 151.08   100
#   never 35.10 37.12 40.89  38.25  42.84  61.31   100
#    once 35.36 37.38 41.58  38.40  44.17  68.43   100
```

R/Spectra.R
```diff
@@ -321,6 +321,27 @@ NULL
...
+#'   peaks of the two spectra that are compared. The function needs to take two
+#'   numeric vectors as input and is supposed to return a single numeric as
+#'   result. Additional parameters to the function can be passed with `...`.
+#'   Prior to the comparison of the intensities, `compareSpectra` matches peaks
+#'   based on the similarity of their m/z with arguments `tolerance` and `ppm`
+#'   allowing to define the maximal allowed difference of m/z values for peaks
+#'   to be matched: `tolerance` allows to define a constant maximal accepted
+#'   difference and `ppm` a relative, m/z dependent, difference (i.e.
+#'   differences smaller than `ppm` parts-per-million of the actual peak's m/z
+#'   are allowed; see [closest()] for more information).
```
sgibb: You should mention here that the real `ppm` is added to `tolerance` so the *real* tolerance per peak is `tolerance + ppm[i]`.

R/Spectra-functions.R
```diff
@@ -153,3 +153,104 @@ applyProcessing <- function(object, f = dataStorage(object),
...
+#' res <- .compare_spectra(sps, sps[1])
+#' @noRd
+.compare_spectra <- function(x, y = NULL, FUN = cor, tolerance = 0,
+                             ppm = 20, ...) {
+    x_idx <- seq_along(x)
+    y_idx <- seq_along(y)
+    mat <- matrix(NA_real_, nrow = length(x_idx), ncol = length(y_idx),
+                  dimnames = list(spectraNames(x), spectraNames(y)))
+    ## Might need some tuning - bplapply?
+    for (i in x_idx) {
```
sgibb: Do we really want to compare each spectrum against each other twice (I know that `.peaks_compare_intensities(x, y)` could be different to `.peaks_compare_intensities(y, x)` depending on the output of `closest`). Here we could avoid half of the computation by just copying the upper triangle of the matrix to the lower (and don't calculate the diagonal at all).
I guess `combn` could help here.

R/Spectra.R
```diff
@@ -321,6 +321,27 @@ NULL
...
+#'   The function matches peaks for a spectrum in `x` with peaks
```
sgibb: This sentence doesn't really make sense here? Is there something missing?

R/Spectra.R
```diff
@@ -1209,7 +1250,33 @@ setMethod("clean", "Spectra",
...
+#'
+#' @exportMethod compareSpectra
+#'
+#' @export ppm closest
+setMethod("compareSpectra", signature(x = "Spectra", y = "Spectra"),
+          function(x, y, FUN = cor, tolerance = 0, ppm = 20, ...,
+                   SIMPLIFY = TRUE) {
+              mat <- .compare_spectra(x, y, FUN = FUN, tolerance = tolerance,
+                                      ppm = ppm, ...)
+              if (SIMPLIFY && (length(x) == 1 | length(y) == 1))
```
sgibb: Doesn't really matter here but would be possible:
```suggestion
              if (SIMPLIFY && (length(x) == 1 || length(y) == 1))
```

R/hidden_aliases.R
```diff
@@ -1,6 +1,6 @@
 #' @title Internal page for hidden aliases
 #'
-#' @aliases [,MsBackendDataFrame-method
+#' @aliases [,MsBackendDataFrame-method ppm closest
```
sgibb: Do we want to export `closest`? Why?

R/peaks-functions.R
```diff
@@ -125,3 +127,66 @@ NULL
...
+#' Spectra:::.peaks_compare_intensities(x, y, FUN = function(x, y) length(x))
+#' Spectra:::.peaks_compare_intensities(x, y, FUN = function(x, y) length(x), ppm = 40)
+#' ## NA if there are none common
+#' Spectra:::.peaks_compare_intensities(x, y, FUN = function(x, y) length(x), ppm = 0)
+.peaks_compare_intensities <- function(x, y, FUN = cor, tolerance = 0,
+                                       ppm = 20, ...) {
+    matches <- closest(x[, 1], y[, 1], tolerance = tolerance + ppm(y[, 1], ppm))
+    not_na <- !is.na(matches)
+    if (any(not_na)) {
+        FUN(x[not_na, 2], y[matches[not_na], 2], ...)
```
sgibb: I am not sure that this function really does what it is intended to do.
See this toy example with just 2 common peaks out of 20 gets a perfect `cor` of `1` with binning it looks completely different:
```r
x <- cbind(1:10, 1:10)
y <- cbind(9:20, 9:20)

.peaks_compare_intensities(x, y)
# [1] 1

xb <- bin(x[, 2], x[, 1], size=1, breaks=1:20)
yb <- bin(y[, 2], y[, 1], size=1, breaks=1:20)
xb <- cbind(xb$mids, xb$x)
yb <- cbind(yb$mids, yb$x)

.peaks_compare_intensities(xb, yb)
# [1] -0.552357
```

R/peaks-functions.R
```diff
@@ -125,3 +127,66 @@ NULL
...
+#' Spectra:::.peaks_compare_intensities(x, y, ppm = 0)
+#'
+#' ## To get the number of common peaks
+#' Spectra:::.peaks_compare_intensities(x, y, FUN = function(x, y) length(x))
+#' Spectra:::.peaks_compare_intensities(x, y, FUN = function(x, y) length(x), ppm = 40)
+#' ## NA if there are none common
+#' Spectra:::.peaks_compare_intensities(x, y, FUN = function(x, y) length(x), ppm = 0)
+.peaks_compare_intensities <- function(x, y, FUN = cor, tolerance = 0,
+                                       ppm = 20, ...) {
+    matches <- closest(x[, 1], y[, 1], tolerance = tolerance + ppm(y[, 1], ppm))
```
sgibb: Shouldn't we use `closest(x[, 1], y[, 1], ... , duplicates="closest")` here to avoid duplicated matches to `y`?

R/AllGenerics.R
sgibb: Don't struggle with this. I will try it myself (because the dispatching happens in `.peaks_bin` actually) once the PR is merged. So ignore it for now.

DESCRIPTION
jorainer: With that we would not need to specifically install `MsCoreUtils` in travis I guess. Ultimately, when all packages are in Bioconductor, we don't need that.

R/AllGenerics.R
jorainer: OK, but yes, that makes sense.

R/Spectra.R
jorainer: Yep, will do.

R/Spectra.R
jorainer: Thanks! yes, makes no sense. will remove.

R/Spectra-functions.R
jorainer: I don't think we are calculating things twice. `x` and `y` are independent vectors (presumably not containing the same spectra). We are comparing each element in `x` with each element in `y`, but not *vice-versa*.

There is a second function for *self*comparison (when `y` is missing) that uses `combn` and copying the lower triangle.

R/peaks-functions.R
jorainer: Yes, makes sense. Will change that.

R/hidden_aliases.R
jorainer: Well, we don't have to. Will remove.

R/peaks-functions.R
jorainer: That's a good point. We should discuss this also in issue #46 - and eventually involve people that do actually MS2 spectra comparisons - I've never done that so far...

jorainer: Binning assumes that the peaks that are not there have an intensity of `0` - my gut feeling tells me that is also not OK.

DESCRIPTION
sgibb: Mh, ok. I already added `MsCoreUtils` to `.travis.yml` just to be sure we always use the latest version. So I would like to remove this line so we can't forget it later.

R/Spectra-functions.R
sgibb: Ok, I missed that. I thought `x` and `y` can contain the same spectra (but they can if the user doesn't take care of it). The comparison of `x[i]` vs `y[j]` is independent so we can do it in parallel. It is not part of the processing queue so we have to do it "manually" here, or?

R/Spectra.R
sgibb: BTW: should be 
```suggestion
#'   are allowed; see [`closest()`] for more information).
```
here, or? (code ticks around closest)

R/peaks-functions.R
sgibb: Code point to ask people that are more familiar with this specific topic. Let's wait what kind of input we get.

R/Spectra.R
jorainer: I removed that link because `closest` is now no longer be exported.

DESCRIPTION
jorainer: OK. Fine with me.

R/Spectra-functions.R
jorainer: Yep, we have to do it manually. I will try but my guess is that parallel processing will not be faster because of the required splitting/merging of the input.

R/Spectra.R
sgibb: Correct.

jorainer:
Closing pull request as we will come up with some better approaches here.

PR #50, opened 2019-07-29
import MsCoreUtils

lgatto:


DESCRIPTION
```diff
@@ -35,6 +36,8 @@ Suggests:
     rhdf5,
     rmarkdown,
     magrittr
+Remotes:
```
jorainer: Actually, @sgibb and I opted some time ago against the use of a `Remote`. During development we will anyway have to have `MsCoreUtils` installed locally and later, in Bioconductor, we don't need it anymore. Having it now might increase the chances to forget removing it later.

R/Spectra.R
```diff
@@ -424,6 +424,8 @@ NULL
 #'
 #' @exportMethod Spectra
 #'
+#' @import MsCoreUtils
```
jorainer: I'm not so happy with importing the whole `MsCoreUtils` namespace - I would rather selectively import only functions needed, and have that import in the header of the function that is using it.

DESCRIPTION
```diff
@@ -24,7 +24,8 @@ Depends:
     ProtGenerics (>= 1.17.2)
 Imports:
     methods,
-    IRanges
+    IRanges,
+    MsCoreUtils
```
sgibb: I am wondering why this is even possible because I add `MsCoreUtils` in https://github.com/rformassspectrometry/Spectra/commit/6518dd8a275f7f8c894f882ce20d4ae3a7ec1419 already.

lgatto:
We need to get the packages to build and check on Travis, be it only to get the webpages up and running. So I think we don't really have the choice of using *Remotes* until the packages are on Bioconductor.

As for imports, I needed to add this as a fix. Selectively importing the functions is indeed a better choice in the future.

R/Spectra.R
lgatto: Yes, selective importing relevant functions is better. I needed a quick fox to get the webpage to build.

DESCRIPTION
lgatto: Sorry if I missed it.

sgibb:
Therefor `Remotes:` in `DESCRIPTION` isn't needed. I added https://github.com/rformassspectrometry/Spectra/blob/c7c8840f2b966d38cb502f85d6d4c49a68e46aca/.travis.yml#L31-L32
in `.travis.yml` which created the webpage before your PR: https://travis-ci.org/rformassspectrometry/Spectra/jobs/560884481

lgatto:
Using `Remotes` of setting the url in the `travis` file is essentially the same. I tend to look at the `DESCRIPTION` file more often, hence easier to check/remember there.

sgibb:
But it has to stay in `.travis.yml` forever because we always want to use the very latest development version.

lgatto:
Indeed.

Issue #49, opened 2019-07-23
align/match peaks of two `Spectrum` objects


tnaake:
Besides the possibility to use `binning` to match spectra, another possibility to align/match peaks of two `Spectrum` objects could be the following using bipartitte networks: 

 1.  start with a bipartite network with all possible edges between `Spetrum` 1 and 2 
 2. remove edges between peaks (1-5, a-f) that fall not within the given `ppm` range spanned by two peaks of `Spectrum` 1 and 2
 3. remove edges between peaks (1-5, a-f) that violate the condition of m/z order maintenance (this step reduces the number of combinations in the next step) 
 4. from the resulting network, obtain all combinations of alignment/matching
 5. calculate the similarity score using a specified `metric` (e.g. NDP) and select the alignment/matching that optimizes the score, calculate average from the different combinations or similar

required input: `Spectrum` object 1, `Spectrum` object 2, `ppm` value (numeric), `metric` (character or a function)

![align_match_spectra](https://user-images.githubusercontent.com/7208089/61729717-23370580-ad78-11e9-87c9-ba29a3215d01.jpg)

jorainer:
I had a thought about the function definition: it should take two arguments `x` and `y`, each being a `matrix` with two columns (first column `"mz"` and second `"intensity"`). The return object should *again* be `list` with elements `x` and `y`, each being a `matrix` (columns `"mz"` and `"intensity"`), but with the same number of rows and each row (peak) in `x` matching the row (peak) in `y`.

I would then use the same definition in the other functions (*join* and *bin*). Any thoughts @lgatto and @sgibb?

tnaake:
Hi @jorainer, @lgatto and @sgibb 

I implemented the aligning/matching as outlined above in the following script that can be found here 
https://github.com/tnaake/R_script/blob/master/match_spectra.R

The function got quite lengthy. Maybe you have some suggestions on improving it. There was one issue with `Spectrum2` objects that required a work-around since Spectrum2 reorders m/z values internally (see within the function for further information). 

Could you please tell me, how I should proceed with it?

jorainer:
Hi @tnaake, thanks for the contribution.

I would suggest we go step-by-step from here. Firstly, please make your code not depend on any objects such as `MSnbase::Spectrum2` or `Spectra` (from `MSnbase` or `Spectra`).

[...]

PR #53, opened 2019-07-31
Add joinPeaks and compareSpectra

jorainer:
This PR adds:

- `joinPeaks` function to map/match peaks from two spectra. It uses the `MsCoreUtils::join` function and could be used as a template for other spectra mapping/matching functions (see also issue #46 and #49).
- `compareSpectra` method to compare spectra between two `Spectra` objects, or witin the same `Spectra` object. Users can configure both the *mapping* function that should be used to map/match the spectra (parameter `MAPFUN`) and the *comparison* function that compares the intensities of the matched spectra (parameter `FUN`).

At present the default for `FUN` is `cor`, which is a poor choice and we might want to use the `dotproduct` instead (will have to be moved to `MsCoreUtils` first).


R/Spectra-functions.R
```diff
@@ -153,3 +153,109 @@ applyProcessing <- function(object, f = dataStorage(object),
...
+    x_idx <- seq_along(x)
+    cb <- combn(x_idx, 2, function(idx) {
+        peak_map <- MAPFUN(peaks(x[idx[1]])[[1]], peaks(x[idx[2]])[[1]],
+                           tolerance = tolerance, ppm = ppm)
+        FUN(peak_map$x[, 2], peak_map$y[, 2], ...)
+    })
+    mat <- matrix(NA_real_, length(x_idx), length(x_idx),
+                  dimnames = list(spectraNames(x), spectraNames(x)))
+    mat[lower.tri(mat)] <- cb
+    for (i in seq_len(nrow(mat))) {
```
sgibb: You could replace the `for` loop by:
```r
# copy lower.tri (needs t and upper.tri because of colum-wise filling) to upper.tri
mat[upper.tri(mat)] <- t(mat)[upper.tri(mat)]
# the comparison function should return the same results if x and y are the same
# so we can just calculate it once and copy it to the whole diagonal
p <- peaks(x[1L])[[1L]][,2L]
diag(mat) <- FUN(p, p, ...)
```

R/Spectra.R
```diff
@@ -1174,23 +1213,23 @@ setMethod("filterRt", "Spectra",
...
+        return(x)
     if (!length(breaks)) {
-        mzr <- range(.peaksapply(filterMsLevel(object, msLevel.),
+        mzr <- range(.peaksapply(filterMsLevel(x, msLevel.),
                                  function(z, ...) z[c(1L, nrow(z))]
                                  ), na.rm = TRUE)
         breaks <- seq(floor(mzr[1]), ceiling(mzr[2]), by = binSize)
     }
-    object <- addProcessing(object, .peaks_bin, breaks = breaks,
+    x <- addProcessing(x, .peaks_bin, breaks = breaks,
```
sgibb: Just coding style: please correct the indentation below.

R/Spectra.R
```diff
@@ -1174,23 +1213,23 @@ setMethod("filterRt", "Spectra",
...
                             msLevel = msLevel.)
-    object@processing <- .logging(object@processing,
+    x@processing <- .logging(x@processing,
```
sgibb: Just coding style: please correct the indentation below.

R/Spectra-functions.R
sgibb:
```r
p <- peaks(x[1L])[[1L]][,2L]
diag(mat) <- FUN(p, p, ...)
```
is not working for comparison functions that e.g. count common peaks.
So may keep the `for` loop but take the copying out of it, or use `vapply1d` for the diagonal: `diag(mat) <- vapply1d(...)`.
It's up to you.

sgibb:
We should provide an easy way to determine the similarity based on common peaks (which is the default in `MSnbase`, currently). While it is easily possible if you know how `compareSpectra` works internally I think it is to difficult for the user:
```r
compareSpectra(sps, type = "left", FUN=function(x, y, ...)length(y))
```

R/Spectra-functions.R
```diff
@@ -153,3 +153,109 @@ applyProcessing <- function(object, f = dataStorage(object),
...
+    x_idx <- seq_along(x)
+    y_idx <- seq_along(y)
+    mat <- matrix(NA_real_, nrow = length(x_idx), ncol = length(y_idx),
+                  dimnames = list(spectraNames(x), spectraNames(y)))
+    ## Might need some tuning - bplapply?
+    for (i in x_idx) {
+        for (j in y_idx) {
+            peak_map <- MAPFUN(peaks(x[i])[[1]], peaks(y[j])[[1]],
+                               tolerance = tolerance, ppm = ppm, ...)
+            mat[i, j] <- FUN(peak_map$x[, 2], peak_map$y[, 2], ...)
```
sgibb: It is not a big deal but `peak_map[[1L]][, 2L]` would be slightly faster than `peak_map$x[, 2]`:
```r
library("microbenchmark")
l <- list(x = 1:1e4, y = 1:1e4)
microbenchmark(l$x, l[[1L]], l$y, l[[2L]])
Unit: nanoseconds
    expr min    lq   mean median    uq  max neval cld
     l$x 217 231.0 355.83  249.5 268.5 9889   100   a
 l[[1L]] 157 172.0 206.36  178.5 188.0 2669   100   a
     l$y 230 247.5 280.07  267.5 285.0  912   100   a
 l[[2L]] 150 165.5 182.89  177.5 197.5  304   100   a
```

R/Spectra-functions.R
```diff
@@ -153,3 +153,109 @@ applyProcessing <- function(object, f = dataStorage(object),
...
+#' all.equal(res, res_2)    # comparison x[1], x[2] != x[2], x[1]
+#' all.equal(res[!lower.tri(res)], res_2[!lower.tri(res_2)])
+#'
+#' @noRd
+.compare_spectra_self <- function(x, MAPFUN = joinPeaks, tolerance = 0,
+                                  ppm = 20, FUN = cor, ...) {
+    x_idx <- seq_along(x)
+    cb <- combn(x_idx, 2, function(idx) {
+        peak_map <- MAPFUN(peaks(x[idx[1]])[[1]], peaks(x[idx[2]])[[1]],
+                           tolerance = tolerance, ppm = ppm)
```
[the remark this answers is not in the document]
jorainer: Yes, thanks!

R/Spectra-functions.R
```diff
@@ -153,3 +153,109 @@ applyProcessing <- function(object, f = dataStorage(object),
...
+#'
+#' @examples
+#'
+#' library(Spectra)
+#' fl <- system.file("TripleTOF-SWATH/PestMix1_SWATH.mzML", package = "msdata")
+#' sps <- Spectra(fl, source = MsBackendMzR())
+#'
+#' sps <- sps[1:10]
+#' res <- .compare_spectra(sps, sps, ppm = 20)
+#' res <- .compare_spectra(sps, sps, method = "spearman", ppm = 40)
```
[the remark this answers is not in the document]
jorainer: Yep, you get `NA` because of the `cor` and the missing `use`. Good point, I will add it.

R/Spectra-functions.R
jorainer: Nice, thanks. Only, I think I can not avoid the for loop: `p <- peaks(x[1L])[[1L]][, 2L]` would only extract the intensities from the first spectrum. I could extract all with `peaks(x)` but would then need a `lapply` or similar already. Instead of `vapply` I would then have to use `mapply` to support functions that calculate something on two numeric vectors (user could provide their own defined functions). Also we would need to pass additional parameters `...` along. In the end I think the `for` loop is easier to understand and should not be much slower than `mapply`.

So, I would use your suggestion with the `upper.tri` and keep the `for` loop for the diagonal.

R/Spectra.R
jorainer: Will do.

R/Spectra.R
jorainer: Will do

R/Spectra-functions.R
jorainer: OK, I will change it.

jorainer:
Regarding the common peaks: yes, we should implement a `commonPeaks` function.

PR #55, opened 2019-08-21
Update to ProtGenerics 1.17.4

jorainer:


lgatto:
version 1.17.4 isn't available on Bioconductor yet. I have temporarily set `lgatto/ProtGenerics` in `MSnbase`s travis file to avoid failing for a couple of days.

Issue #57, opened 2019-08-23
change filterPrecursorMz to accept an m/z range


jorainer:
The current implementation of `filterPrecursorMz` has parameters `mz` and `ppm` both being a `numeric(1)`. This is however different from the way `filterMz` is used (i.e. takes an `m/z` of length 2) which I find confusing.

I'd propose to change `filterPrecursorMz` to take also an `mz` being a `numeric(2)` defining the lower and upper m/z boundary and removing the `ppm` parameter alltogether.

Any comments @lgatto @sgibb ?

lgatto:
Yes for consistency.

I have one thought though. Setting the range by hand as you suggest is more convenient, but this range isn't equivalent throughout the whole m/z range (this [question](https://www.biostars.org/p/8453/) summarises my thoughts). Using an m/z of `length` 2 wouldn't necessary be ideal to filter specific peaks at very different m/z. Or may be it doesn't matter that much?

jorainer:
I'm a little confused by the thoughts you expressed. 
Regarding the `ppm`, I thought you use a relative accuracy like that because TOF MS are less accurate for higher masses but I could be mistaken here.

> Using an m/z of length 2 wouldn't necessary be ideal to filter specific peaks at very different m/z. 

Why? If you provide `mz = c(123.2, 123.3)` you would get all spectra with a precursor m/z within that range - if you would want to use a single m/z with a ppm of 10 you could do `mz = 123.2 + ppm(c(-123.3, 123.2), 10)`. 

What would not be possible (but that was also not possible before) was to filter for multiple m/z simultaneously, i.e. get all spectra that have a precursor m/z of 123.3 **or** 123.6 **or** 343.2 with a **single** `filterPrecursorMz` call.

lgatto:
I was referring to a case where one would want to filter out a peak at a low and a high m/z range. `mz = 123.2 + ppm(c(-123.3, 123.2), 10)` solves it, thanks.

jorainer:
I will add this also as an example.

Issue #56, opened 2019-08-21
Import bin from MsCoreUtils


jorainer:
Import the `bin` function from `MsCoreUtils` and remove the `bin_values` function.

PR #58, opened 2019-08-23
Remove parameter ppm from filterPrecursorMz and code cleanup

jorainer:
- Remove the `ppm` parameter from the `filterPrecursorMz` function that expects now a `numeric(2)` with parameter `mz` (issue #57).
- Remove no longer needed binning utility functions (issue #56).


jorainer:
No idea why I get now the check error - will investigate and fix.

jorainer:
Hm `S4Vectors::DataFrame` returns now a `DFrame` instead of a `DataFrame`!

```r
> class(S4Vectors::DataFrame())
[1] "DFrame"
attr(,"package")
[1] "S4Vectors"

```

lgatto:
Indeed, but only in devel. That is confusing. I don't remember seeing this on the developer mailing list. It's very recent:

```
Author: Hervé Pagès <hpages@fredhutch.org>
Date:   Tue Aug 20 19:02:00 2019 -0700

    DataFrame() now returns a DFrame instance
    
    Other changes:
    
    o All coercion methods to DataFrame now return a DFrame instance instead
      of a DataFrame instance. The goal being that, from now on, any code that
      used to produce a DataFrame instance will instead produce a DFrame
      instance. This is achieved by having the DataFrame() constructor and all
      the coercion methods to DataFrame return a DFrame instance instead of a
      DataFrame instance.
    
    o show() still prints a header line that says "DataFrame with M rows and
      N columns" on a DFrame instance. This is in order to make the migration
      from DataFrame to DFrame as transparent as possible for the end-user
      (although s/he'll notice the change when calling class() on the object).
    
    o Calling updateObject() on a DataFrame instance will turn it into a DFrame
      instance. Also calling it on a Vector derivative that has its metadata
      columns in a DataFrame instance will return an object where the metadata
      columns are in a DFrame instance. This is to facilitate the migration of
      serialized DataFrame instances or serialized Vector derivatives containing
      DataFrame instances.
    
    o 'mcols(x)' and 'mcols(x) <- value' now pass 'x@elementMetadata' and
      'value' (respectively) thru updateObject(). This reduces the risk of
      propagating DataFrame instances by automatically replacing them
      on-the-fly with DFrame instances.

commit 36837bdf8ba55a3e12f8f55cb3cf75c45b045d77
Author: Hervé Pagès <hpages@fredhutch.org>
Date:   Fri Aug 16 08:20:25 2019 -0700

    Add DFrame class
    
    Just a direct DataFrame extension with no additional slot for now. Once all
    serialized DataFrame instances are replaced with DFrame instances (which
    will take several BioC release cycles) we'll be able to move the DataFrame
    slots from the DataFrame class definition to the DFrame class definition.
    The final goal is to have DataFrame become a virtual class with no slots
    that only extends DataTable, and DFrame a concrete DataFrame and SimpleList
    subclass that has the same slots as the current DataFrame class.
```

lgatto:
For the record, @hpages announcing the [`DFrame` change](https://stat.ethz.ch/pipermail/bioc-devel/2019-August/015416.html) on bioc-devel. This is actually excellent news for our back-end work.

lgatto:
For the record, @hpages announcing the [`DFrame` change](https://stat.ethz.ch/pipermail/bioc-devel/2019-August/015416.html) on bioc-devel. This is actually excellent news for our back-end work.

PR #63, opened 2019-09-10
Some refactoring

jorainer:
- Replace `.i_to_index` with `i2index` from `MsCoreUtils`.
- Remove `validObject(object)` calls from the `[` methods of the backends. Reason: considerable speed improvement for e.g. `split` calls etc. IMHO validity checks are not required for subsetting operations - in contrast to e.g. replacement operations.


jorainer:
What you think about removing the `validObject` calls in `[` @sgibb ? Are you OK with that?

sgibb:
I already forgot that. (I was so focused on the `i2index` and `MsCoreUtils` *master* vs *devel* problem.)
In principle I don't really like that. How much time we save? Would it be possible to create a `.subset` function (without validation) that could be reused in `[` and `split` (so we avoid duplicated `validObject` calls)?

jorainer:
I'll have a look at the `[` and `split` methods in `IRanges`/`S4Vectors`. They have an incredibly fast `split` function there (0.001 seconds on a plain `DataFrame` versus 6.2 seconds for a `split` on one of our backends - after removing the `validObject` - with `validObject` it was 22 seconds).

The `validObject` will slow us down also in the `compareSpectra` where we loop over all spectra and in each loop subset with `x[i]`. The problem is if the `Spectra` contains thousands of spectra, then it becomes slow.

sgibb:
Ok 6 vs 22 is a real benefit. So go ahead. What is the magic in `S4Vectors::split`?

jorainer:
Re magic in `S4Vectors` - I'll have to check - let's see if we can adapt something from there.

Issue #61, opened 2019-09-05
Define an endoapply and lapply method


jorainer:
Implement an `lapply` and `endoapply` method (internal function) to apply any function on a `Spectra` object (of length 1).

There is already the internal `.peaksapply` that applies a function to the *peaks* matrix. An updated version of this function should take with `x` a `matrix` with the m/z-intensity pairs and a `list` (`List`) with the spectra variables.

jorainer:
We might have to implement a `split` method for `Spectra`/`MsBackend` as the *default* one seems to be relatively slow.

jorainer:
Some info on the `split` method for `DataFrame`: calling `split` on a `DataFrame` is **very** fast, but does in fact **not** split the data - it returns a `CompressedSplitDFrameList` (which, as far as I understand, is still the original `DataFrame` just with partitioning information). Now, running `lapply` on such a *virtually* splitted data is actually slow. A call to `lapply` to a `list` is way faster.

The `base::split` function is actually *relatively* fast, I don't think we can make that one faster. It is also faster then calling `split,DataFrame` followed by `as.list` on the `CompressedSplitDFrameList` (which will in fact return the `list` with individual `DataFrame` objects).

So, in the end I don't think we need to implement a dedicated `split` method for `Spectra` or any `MsBackend`, but we have to ensure that the `[` method is as fast as possible.

jorainer:
Closing the issue, since `lapply` is implemented.

PR #65, opened 2019-09-13
feat: Add `lapply,Spectra`

jorainer:
Add a `lapply,Spectra` method to apply any function to each spectrum of a `Spectra` object (issue #61).


R/MsBackendDataFrame-functions.R
```diff
@@ -217,3 +217,28 @@ MsBackendDataFrame <- function() {
...
+#' @importMethodsFrom S4Vectors extractROWS
+#'
+#' @importFrom methods slot<-
+#'
+#' @noRd
+.subset_backend_data_frame <- function(x, i) {
+    if (missing(i))
+        return(x)
+    i <- i2index(i, length(x), rownames(x@spectraData))
+    slot(x, "spectraData", check = FALSE) <- extractROWS(x@spectraData, i)
```
sgibb: I guess slot(..., check = FALSE) <- is faster than @... because of the avoided checks?

jorainer: It avoids checking that the `value` is of the correct type. That's how it's done in `S4Vectors` that's why I've used it - but I haven't seen any major speed improvement.

Issue #67, opened 2019-09-16
MsBackendMzR: don't import core spectra variables with only NA


jorainer:
The `MsBackendMzR` backend should not import core spectra variables (columns from `mzR::header`) if they contain only `NA` values. These would be added as `Rle` (containing only `NA`) that can affect/slow down subsetting with `[` and `split`.

Issue #44, opened 2019-06-24
Bypass .peaksapply if processing queue is empty


jorainer:
For performance reasons, we might want to bypass the `.peaksapply` function if the processing queue is empty for the `mz,Spectra`, `intensity,Spectra` and `peaks,Spectra` methods.

jorainer:
Update on this. Some timings for `mz,Spectra` (on an in-memory or on-disk backend) and the corresponding direct call to the backend's `mz` method. `mz,Spectra` calls `.peaksapply`, which splits the object by `dataStorage` and calls `peaks` in parallel (eventually applying the lazy processing queue). `mz,Spectra` is thus expected to be slower. The example below based on 16 mzML files

```r
library(Spectra)
library(microbenchmark)
register(bpstart(MulticoreParam(4)))

## Eval on a data set with 16 files.
fls <- dir("/Users/jo/data/2018", pattern = "mzML", full.names = TRUE)
sps_mzr <- Spectra(backendInitialize(MsBackendMzR(), fls[1:16]))
sps_im <- setBackend(sps_mzr, MsBackendDataFrame())

microbenchmark(mz(sps_mzr),
               mz(sps_mzr@backend),
               mz(sps_im),
               mz(sps_im@backend),
               times = 3
               )
## Unit: microseconds
##                 expr          min           lq         mean       median
##          mz(sps_mzr) 36470168.349 47724429.547 52754600.174 58978690.745
##  mz(sps_mzr@backend) 27829000.577 30114077.483 32931790.305 32399154.389
##           mz(sps_im)  2736414.926  3512628.783  4667086.390  4288842.639
##   mz(sps_im@backend)      357.322      365.754      377.476      374.186
##            uq         max neval cld
##  60896816.087 62814941.43     3   b
##  35483185.169 38567215.95     3   b
##   5632422.122  6976001.61     3  a
##       387.553      400.92     3  a
```

Firstly, in-memory backend is faster than the on-disk version. Secondly, the direct call to the backend outperforms the `.peaksapply`-based one - very clearly for the in-memory backend, but, somewhat surprisingly, also for the on-disk backend. Most likely because reading spectra from the mzML files is faster than the splitting and merging required by the parallel processing.

jorainer:
For very large experiments/`Spectra` there might be a performance loss for on-disk objects - that is however outweighted by the very large performance gain of in-memory backends.

PR #71, opened 2019-09-26
refactor: changes addressing performance bottlenecks

jorainer:
- `mz,Spectra` and `intensity,Spectra` avoid parallel processing. This significantly improves performance on in memory backends (issue #44).
- Avoid import of core spectra variables with only `NA` in `MsBackendMzR` (issue #67).

### DIFF
--- a/R/MsBackendMzR-functions.R
+++ b/R/MsBackendMzR-functions.R
@@ -15,6 +15,8 @@ MsBackendMzR <- function() {
 #'
 #' @author Johannes Rainer
 #'
+#' @importFrom MsCoreUtils vapply1l
+#'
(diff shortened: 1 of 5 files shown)


jorainer:
Wow! That was a fast review @sgibb! And no comments? Sure?

sgibb:
@jorainer it was a short PR as well. I am sure everything is fine. So go ahead and merge.

BTW: In general all these special treatments for performance reasons are causing me a little bit headache because I am afraid we lose maintainability.

jorainer:
I agree - we should try to avoid these, but in this case I had the feeling that it might be OK...

Issue #62, opened 2019-09-09
Replace peaks method by as.list


jorainer:
Replace the `peaks` method by `as.list` (see https://github.com/rformassspectrometry/RforMassSpectrometry/issues/14).

jorainer:
Here we run into a problem: replacing `peaks` by `as.list` would be OK, but we have also a `peaks<-` method that allows to replace the m/z and intensity values. I think it would be crucial to have a dedicated method to replace the m/z-intensity **pairs**, especially because `mz<-` or `intensity<-` would not work because they would have to be called sequentially with the first potentially leaving the data in an invalid state (e.g. because replacing m/z values first will result in a different number of m/z and intensity values).

I would thus propose to add a `mzIntensityPairs` and `mzIntensityPairs<-` method (replacing the `peaks` method) and making a symlink from `as.list` to `mzIntensityPairs`. I know, the name is pretty ugly, but we could then have the analogues in `Chromatograms`: `rtimeIntensityPairs` and `rtimeIntensityPairs<-`.

What do you think @lgatto @sgibb ?

sgibb:
Or `replaceList` (in analogy to `S4Vectors::replaceCOLS`/`S4Vectors::replaceROWS`) or `list2spectra`/`list2chromatograms` (in analogy to `base::list2env`/`base::list2DF`)?

Or we define our own `as.list<-` operator (`as.list(object) <- value`)

jorainer:
Excellent idea @sgibb! I like the `replaceList<-`! (I find the `as.list<-` a little unintuitive).

lgatto:
Ok, that seem sensible to me.

PR #72, opened 2019-09-27
refactor: replace `peaks` method with `as.list` (issue #62)

jorainer:
- Replace `peaks` with `as.list` and `peaks<-` with `replaceList`.
- Update documentation and unit tests.


R/MsBackend.R
```diff
@@ -125,6 +125,12 @@ NULL
 #'   spectrum. Returns an `integer` of length equal to the number of
 #'   spectra (with `NA_integer_` if not available).
 #'
+#' - `as.list` returns a `list` with the spectras' peak data. The length of
+#'   the list is equal to the number of spectra in `object`. Each element of
+#'   the list is a `matrix` with columns `mz` and `intensity`. For an empty
```
sgibb: Here you use:
> with columns `mz` and `intensity`

and for `replaceList` you use

> with columns `"mz"` and `"intensity"`

Please keep one of them. (I think the `"column"` version is easier to understand.)

R/MsBackend.R
```diff
@@ -287,6 +283,12 @@ NULL
 #'   the number of spectra in `object`. `NA` are reported for MS1
 #'   spectra of if no precursor information is available.
 #'
+#' - `replaceList<-` replaces the peak data (m/z and intensity values) of the
+#'   backend. This method expects a `list` of `matrix` objects with columns
+#'   `"mz"` and `"intensity"` that has the same length than the number of
```
sgibb: See comment above.

sgibb:
```suggestion
#'   `"mz"` and `"intensity"` that has the same length as the number of
```

R/MsBackendHdf5Peaks.R
```diff
@@ -157,7 +175,7 @@ setMethod("isEmpty", "MsBackendHdf5Peaks", function(x) {
 
 #' @rdname hidden_aliases
 setMethod("mz", "MsBackendHdf5Peaks", function(object) {
-    NumericList(lapply(peaks(object), "[", , 1), compress = FALSE)
+    NumericList(lapply(as.list(object), "[", , 1), compress = FALSE)
```
sgibb: This `list(lapply(as.list(` statement looks great now :wink:

R/MsBackendHdf5Peaks.R
```diff
@@ -172,30 +190,16 @@ setReplaceMethod("mz", "MsBackendHdf5Peaks", function(object, value) {
...
-        unsplit(bpmapply(
-            FUN = .h5_read_peaks,
-            fls,
-            split(scanIndex(object), f),
-            object@modCount,
-            SIMPLIFY = FALSE, USE.NAMES = FALSE, BPPARAM = bpparam()),
-            f)
-    } else
-        .h5_read_peaks(fls, scanIndex(object), object@modCount)
+setMethod("peaksCount", "MsBackendHdf5Peaks", function(object) {
```
sgibb: Should this be `lengths` or so? Instead of **peaks**Count?

R/Spectra.R
```diff
@@ -89,6 +89,11 @@ NULL
 #'   spectrum. Returns an `integer` of length equal to the number of
 #'   spectra (with `NA_integer_` if not available).
 #'
+#' - `as.list`: gets the *peaks* matrices for all spectra in `object`. The
+#'   function returns a [SimpleList()] of matrices, each `matrix` with columns
+#'   `mz` and `intensity` with the m/z and intensity values for all peaks of
```
sgibb: See comment about `"mz"` vs `mz` and `"intensity"` vs `intensity` above.

R/MsBackend.R
```diff
@@ -881,6 +874,13 @@ setMethod("precursorMz", "MsBackend", function(object) {
     stop("Not implemented for ", class(object), ".")
 })
 
+#' @exportMethod replaceList
+#'
+#' @rdname MsBackend
+setMethod("replaceList", "MsBackend", function(object, value, ...) {
```
sgibb: Could/Shouldn't this be a `setReplaceMethod` as well?

R/MsBackend.R
jorainer: Yep. will change that. Thanks.

R/MsBackendHdf5Peaks.R
jorainer: Sure! Yes! We said that. I will remove the `peaksCount` method.

R/MsBackend.R
jorainer: I followed what they did in `S4Vectors`. The `replaceROW` and `replaceCOL` are **not** replace methods, they're just simple methods. I have no strong opinion here. If you want me to change I'll do.

R/MsBackendHdf5Peaks.R
jorainer: Hm, only thing is that there's no generic `lengths`.

R/MsBackend.R
sgibb: Me neither. Is there a performance drop? Or is `setReplaceMethod` internally just a synonym for `object <- setMethod`? (I don't have access to R at the moment.)

jorainer: Could one advantage be to have a dispatch on both `object` and `value`?

jorainer: Anyway, I'm going to change it to a replace method. Usage will then be:

```r
lst <- as.list(be)
replaceList(be) <- lst
```

I think that might be easier for users/developers to understand.

jorainer:
Sorry for the many commits. What I did was:
- change `replaceList` to `replaceList<-`.
- fix/unify your comments on documentation.
- replace `peaksCount` by `lengths`.
- Fix unit tests for latest mzR (which caused Travis to fails).

R/MsBackend.R
```diff
@@ -245,6 +253,10 @@ NULL
 #'
 #' - `length`: returns the number of spectra in the object.
 #'
+#' - `lengths`: gets the number of peaks (m/z-intensity values) per
+#'   spectrum.  Returns an `integer` vector (length equal to the
+#'   number of spectra). For empty spectra, `NA_integer_` is returned.
```
sgibb: Mh, does `lengths` really return `NA` for empty spectra? Not zero? Is `NA` a good choice here?

jorainer: Thanks for spotting this! The way it is implemented it would report `0` not `NA`. I'll fix the documentation.

PR #73, opened 2019-10-16
refactor: use dotproduct as default function to compare spectra

jorainer:
This is to adapt to current changes in `MsCoreUtils`, mainly to use the newly added `dotproduct` function:
- Use the `MsCoreUtils::dotproduct` as default for parameter `FUN` of
  `compareSpectra`.
- Adapt the functionality to compare spectra to take the peak matrices as input
  for the comparison function(s).
- Amend documentation and unit tests.


jorainer:
Thanks @sgibb for the fast review!

Issue #74, opened 2019-10-18
Add a `combineSpectra` equivalent method


jorainer:
`MSnbase::combineSpectra` was a function to combine sets of spectra to single `Spectrum` objects.

Now, for `Spectra` I thought we might change that behaviour slightly. I'd propose to have *combining functions* that would reduce a `Spectra` of length `n` to a `Spectra` of length 1 and to have a general function `combineSpectra` that allows to apply any function to sets within `Spectra`.

Functions to combine spectra could be:
- `consensusSpectrum`: same as `MSnbase::consensusSpectrum`
- `union`: combine all spectra into one spectrum keeping all peaks (not sure if this is not redundant to `combinePeaks` below).
- `combinePeaks`: same as the `MSnbase::meanMzInts`, take mean m/z and intensities for peaks within a certain `ppm`.

So, spectra could be combined like that:

```r
## Reduce a Spectra of length x to a Spectra of length 1
sps_1 <- combinePeaks(sps)

## Reduce sets of Spectra within `sps`, defined by column `"group"` to single `Spectra` each
sps_s <- combineSpectra(sps, f = sps$group, FUN = combinePeaks)
```

Any objections/suggestions @lgatto @sgibb ?

PR #75, opened 2019-10-24
Add combineSpectra

jorainer:
Add the `combineSpectra` function for `Spectra` (issue #74 ):

The definition of the function is the following:

```r
combineSpectra <- function(x, f = x$dataStorage, p = x$dataStorage,
                           FUN = combinePeaks, ..., BPPARAM = bpparam())
```

with `x` being `Spectra`. `f` allows to define the grouping of the spectra, i.e. which sets of spectra in `x` should be combined into a single one. I've also added a parameter `p` that allows to define an additional splitting of `x` for parallel processing. By default (with `x$dataStorage`) parallel processing will be performed on a per-file basis (for mzR and Hdf5 backends). This reduces memory demand for such backends and can improve performance. Parameter `FUN` allows to specify a function that combines peaks of spectra.

Such functions are supposed to take a `list` of peak matrices as input and return a single `matrix`.

Note: this uses the `MsCoreUtils::group` function from https://github.com/rformassspectrometry/MsCoreUtils/pull/24


.travis.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         - knitr
         - roxygen2
       r_github_packages:
-        - rformassspectrometry/MsCoreUtils@devel
+        - rformassspectrometry/MsCoreUtils@jomaster
```
sgibb:
```suggestion
        - rformassspectrometry/MsCoreUtils
```
Please change this into `rformassspectrometry/MsCoreUtils` before merging (assuming we merged `group` into `MsCoreUtils`)

R/Spectra-functions.R
```diff
@@ -331,3 +331,99 @@ applyProcessing <- function(object, f = dataStorage(object),
...
+    if (any(cls != "Spectra"))
+        stop("Can only concatenate 'Spectra' objects")
+    pqs <- lapply(x, function(z) z@processingQueue)
+    ## For now we stop if there is any of the processingQueues not empty. Later
+    ## we could even test if they are similar, and if so, merge.
+    if (any(lengths(pqs)))
+        stop("Can not concatenate 'Spectra' objects with non-empty ",
+             "processing queue")
+    metad <- do.call(c, lapply(x, function(z) z@metadata))
+    procs <- unique(unlist(lapply(x, function(z) z@processing)))
```
sgibb: Just a thought: should we provide processing logs in a format like `timestamp - what was done`. Than we could order it here?

R/peak-list-functions.R
```diff
@@ -0,0 +1,189 @@
...
+#'
+#' @details
+#'
+#' For general merging of spectra, the `tolerance` and/or `ppm` should be
+#' manually specified based on the precision of the MS instrument. Peaks
+#' from spectra with a difference in their m/z being smaller than `tolerance`
+#' or smaller than `ppm` of their m/z are grouped into the same final peak.
+#'
+#'
+#' Some details for the combination of consecutive spectra of an LC-MS run:
```
sgibb: Nice explanation :+1:

.travis.yml
jorainer: Jepp. For sure.

R/Spectra-functions.R
jorainer: We could - but I would like to keep things as simple as possible. If so we would have to ensure that a timestamp is always provided...

sgibb: There was this `logging` function I had in mind a few weeks ago https://github.com/rformassspectrometry/MsCoreUtils/pull/5 :wink: but we don't have to do/use that.

jorainer: We can add that later - once the logging functionality is in place

PR #76, opened 2019-10-29
Add Spectra,smooth

sgibb:


R/peaks-functions.R
```diff
@@ -261,3 +261,26 @@ joinPeaks <- function(x, y, type = "outer", tolerance = 0, ppm = 10, ...) {
...
+#' @description
+#'
+#' Intensity smoothing.
+#'
+#' @inheritParams .peaks_remove
+#' @param coef `matrix` smoothing coefficients (generated by
+#' [MsCoreUtils::coefMA()], [MsCoreUtils::coefWMA()], or [MsCoreUtils::coefSG].
+#'
+#' @return `matrix` with columns `"mz"` and `"intensity"`.
+#'
```
jorainer: would you mind adding you here as `@author`? That way we know who to blame if we find a bug in the function ;)

PR #78, opened 2019-12-03
pickPeaks changes centroided status

jorainer:
- Add the `consensusPeaks` function.

- Change `pickPeaks` to set `centroided` to `TRUE` for spectra that will be affected by the centroiding.


R/peak-list-functions.R
```diff
@@ -187,3 +197,76 @@ combinePeaks <- function(x, intensityFun = base::mean,
...
+#' ## peaks
+#' consensusPeaks(list(p1, p2, p3), minProp = 0.2)
+#'
+#' ## With a tolerance of 0.1 mass peaks can be matched across spectra
+#' consensusPeaks(list(p1, p2, p3), tolerance = 0.1)
+#'
+#' ## Report the minimal m/z and intensity
+#' consensusPeaks(list(p1, p2, p3), tolerance = 0.1, intensityFun = min,
+#'     mzFun = min)
+consensusPeaks <- function(x, minProp = 0.5, intensityFun = base::mean,
```
sgibb: The first lines are very similar to `combinePeaks`. Should we create an internal function for them? Here the `timeDomain` argument is missing. Is that intended?

All in all the functions `combinePeaks` and `consensusPeaks` are very similar. Can't we combine them into one?

R/peaks-functions.R
```diff
@@ -239,7 +239,7 @@ joinPeaks <- function(x, y, type = "outer", tolerance = 0, ppm = 10, ...) {
                         halfWindowSize = 2L, method = c("MAD", "SuperSmoother"),
                         snr = 0L, k = 0L, descending = FALSE, threshold = 0,
                         msLevel = spectrumMsLevel, ...) {
-    if (!(spectrumMsLevel %in% msLevel) || isTRUE(centroided))
+    if (!(spectrumMsLevel %in% msLevel))
```
sgibb: `joinPeaks` should work on spectra level as well I guess?

jorainer:
Makes sense - for the use cases:
- `consensusSpectra`: if you have a set of (MS2) fragment spectra from a single precursor m/z and you want to *reduce* this to a singe *representative* spectrum.
- `combineSpectra`: if you want to combine (merge) a set of spectra into one. The original use case was to do a smoothing of signal along retention time. The approach was to use a moving window to *average* signals from each set of consecutive spectra and replace the spectrum in the middle of the window with the merged one. This also explains the *peculiar* special parameters it has (`main`, `unionPeaks`), as we want only to keep peaks that are present in the *middle* (`main`) spectrum.

But you're right. It makes more sense to combine the two functions. We just have to think of clever naming of parameters. I would suggest:
- `peaks = c("union", "intersect")` to switch between `combinePeaks` and `consensusPeaks`.
- rename `unionPeaks` to `keepMain` or `onlyMain`.

Eventually something to discuss briefly during a coffee break at EuroBioc2019?

R/peaks-functions.R
jorainer: haven't thought about that - so you mean to add a parameter `msLevel` to it?

sgibb: I was just asking because you removed the `isTRUE(centroided)` test here. `msLevel` is already present.

jorainer: I had to remove this test, because calling `pickPeaks` will now change `centroided` to `TRUE` immediately. So once we execute the processing queue would not get centroided anymore because `centroided` is already `TRUE`. This was the main problem why I introduced this change.

R/Spectra.R
```diff
@@ -1466,6 +1465,7 @@ setMethod("pickPeaks", "Spectra",
                             halfWindowSize = halfWindowSize, method = method,
                             snr = snr, k = k, descending = descending,
                             threshold = threshold, msLevel = msLevel.)
+    object$centroided[msLevel(object) %in% msLevel.] <- TRUE
```
sgibb: I am still not convinced to set this "undelayed" (immediately). If we chain functions in the processing queue that depend on `centroided == FALSE` before `pickPeaks` they will fail (currently we don't have one but who knows).

Issue #83, opened 2019-12-11
Remove the Rle-based compression


jorainer:
After some discussions @sgibb and I would like to remove the `Rle`-based *compression* that we have by default in the `MsBackendDataFrame`. The reasons for that:

1) memory gain is almost neglectible
2) performance loss is not acceptable considering 1)
3) some optimizations we made make the `Rle` compression obsolete: `split` does uncompress the full `DataFrame` before actually splitting the data (otherwise `split` will be unacceptable slow).

Examples and comparisons follow.

jorainer:
Example:

```r
library(Spectra)
sciex_file <- dir(system.file("sciex", package = "msdata"), full.names = TRUE)
sciex <- Spectra(sciex_file, backend = MsBackendMzR())
```

I'm changing the backend now to the `MsBackendDataFrame` (which has no `Rle` compression) and the `MsBackendDFrame` (which is essentially our old `MsBackendDataFrame` with `Rle` compression).

```r
sciex_df <- setBackend(sciex, MsBackendDataFrame())
sciex_cdf <- setBackend(sciex, MsBackendDFrame())
print(object.size(sciex_df), unit = "MB")
52.4 Mb
print(object.size(sciex_cdf), unit = "MB")
52.1 Mb
```

So, the gain in memory using `Rle` compressionn is limited.

```r
microbenchmark(sciex_df[3:20],
               sciex_cdf[3:20])
Unit: milliseconds
            expr    min      lq     mean  median     uq     max neval cld
  sciex_df[3:20] 1.9891 2.58925 3.081473 2.86740 3.3252  7.7121   100  a 
 sciex_cdf[3:20] 4.0267 5.12880 6.296698 6.10495 7.0004 12.3842   100   b
``` 

So, uncompressing the `Rle` comes at a considerable cost.

PR #85, opened 2019-12-11
Remove all functionality related to Rle-based compression

jorainer:
Remove all functionality for `Rle`-based functionality from `MsBackendDataFrame` (and in fact from the whole package) - as discussed in issue #83 .


jorainer:
Sorry - still having some issues with the tests - should be fixed soon

jorainer:
OK, now it's ready @sgibb if you want to cross-check.

R/MsBackendDataFrame.R
```diff
@@ -97,21 +95,20 @@ setMethod("as.list", "MsBackendDataFrame", function(x) {
...
-#' @importFrom MsCoreUtils asRle
 setReplaceMethod("centroided", "MsBackendDataFrame", function(object, value) {
     value_len <- length(value)
     value_type <- is.logical(value)
     if (value_len == 1 && value_type)
-        object@spectraData$centroided <- Rle(value, length(object))
+        object@spectraData$centroided <- rep_len(value, length(object))
     else if (value_len == length(object) && value_type)
-        object@spectraData$centroided <- asRle(value)
+        object@spectraData$centroided <- value
```
sgibb: Do we really need this `if`/`else` here? `data.frame(...)$column <- value` will replicate the value automatically if `length(value) == 1`.

R/MsBackendDataFrame.R
```diff
@@ -297,17 +294,17 @@ setReplaceMethod("mz", "MsBackendDataFrame", function(object, value) {
...
 #' @rdname hidden_aliases
 setReplaceMethod("polarity", "MsBackendDataFrame", function(object, value) {
     value_len <- length(value)
     value_type <- is.numeric(value)
     if (value_len == 1 && value_type)
-        object@spectraData$polarity <- Rle(as.integer(value), length(object))
+        object@spectraData$polarity <- rep_len(as.integer(value), length(object))
     else if (value_len == length(object) && value_type)
-        object@spectraData$polarity <- asRle(as.integer(value))
+        object@spectraData$polarity <- as.integer(value)
```
sgibb: Same as for `centroid` above.

R/MsBackendDataFrame.R
```diff
@@ -403,9 +400,9 @@ setReplaceMethod("smoothed", "MsBackendDataFrame", function(object, value) {
     value_len <- length(value)
     value_type <- is.logical(value)
     if (value_len == 1 && value_type)
-        object@spectraData$smoothed <- Rle(value, length(object))
+        object@spectraData$smoothed <- rep_len(value, length(object))
     else if (value_len == length(object) && value_type)
-        object@spectraData$smoothed <- asRle(value)
+        object@spectraData$smoothed <- value
```
sgibb: Same as above for `centroid`.

R/MsBackendDataFrame.R
```diff
@@ -529,9 +521,6 @@ setMethod("[", "MsBackendDataFrame", function(x, i, j, ..., drop = FALSE) {
 setMethod("split", "MsBackendDataFrame", function(x, f, drop = FALSE, ...) {
     if (!is.factor(f))
         f <- as.factor(f)
-    if (length(levels(f)) > (nrow(x@spectraData) / 10))
```
sgibb: I always disliked this `if` statement. Now its gone. Yeah! :tada:

R/MsBackendMzR.R
```diff
@@ -177,9 +175,9 @@ setReplaceMethod("$", "MsBackendMzR", function(x, name, value) {
         stop("'MsBackendMzR' does not support replacing mz or intensity values")
     value_len <- length(value)
     if (value_len == 1)
-        x@spectraData[[name]] <- Rle(value, length(x))
+        x@spectraData[[name]] <- rep_len(value, length(x))
     else if (value_len == length(x))
-        x@spectraData[[name]] <- asRle(value)
+        x@spectraData[[name]] <- value
```
sgibb: Same as above for `centroid`.

R/MsBackendDataFrame.R
jorainer: The only reason that I left the `if` `else` here is that I wanted to avoid the automatic replication of R (i.e. if the `DataFrame` has 12 rows, `value` can also be of length 2, 3, 4 or 6 - but I wanted only to support either length 1 or length equal nrow of `DataFrame`).

jorainer:
I agree that your proposed channge would remove some lines of code, but I think we can live with this if we want to allow `value` of either be length 1 or equal nrow of the data frame. I have a small preference to keep the `if`/`else` to avoid the R-inherit replication but am also fine if you insist on removing it.

R/MsBackendDataFrame.R
sgibb: Ok, but then something like the following should be enough, or?
```r
if (value_type && (value_len == 1L || value_len == length(object)))
    object@spectraData$centroided <- value  
```

jorainer: OK. I will adapt.

jorainer: Changed now

Issue #92, opened 2020-03-04
spectraData pulls mz and intenstities


lgatto:
`spectraData` extracts by default the *mz* and *intensities*, which makes the call extremely slow and return a big object:

```r
> basename(f <- msdata::proteomics(pattern  = "20141210", full.names = TRUE))
[1] "TMT_Erwinia_1uLSike_Top10HCD_isol2_45stepped_60min_01-20141210.mzML.gz"
> sp <- Spectra(f, backend = MsBackendMzR())
> system.time(spd1 <- spectraData(sp, setdiff(spectraVariables(sp), c("mz", "intensity"))))
   user  system elapsed 
  0.018   0.000   0.018 
> print(object.size(spd1), units = "Mb")
2.8 Mb
> system.time(spd2 <- spectraData(sp))
   user  system elapsed 
 16.447   0.267  16.571 
> print(object.size(spd2), units = "Mb")
576.1 Mb
```

I would suggest, at least for on-disk backends, that mz and intensity should not be part of the default spectraVariables. In these cases, we could add

```r
spectraData(sp, mz = FALSE, intensity = FALSE)
```

as defaults, that get set to TRUE when these are provided explicitly as part of the `spectraVariables` argument.

Note however that this is dramatically reduced when using the hfd5 backend.

```r
> sp2 <- setBackend(sp, backend = MsBackendHdf5Peaks())
> system.time(spd3 <- spectraData(sp2, setdiff(spectraVariables(sp), c("mz", "intensity"))))
   user  system elapsed 
  0.015   0.000   0.015 
> print(object.size(spd3), units = "Mb")
2.8 Mb
> system.time(spd4 <- spectraData(sp2))
   user  system elapsed 
  4.673   0.188   4.973 
> print(object.size(spd4), units = "Mb")
576.1 Mb
```

@jorainer @sgibb Comments, suggestions?

Cc @plantton

jorainer:
Honestly, I don't like the idea to add `mz` and `intensity` as dedicated parameters to the function. If users want specific columns they can use the `columns` parameter. Also, `spectraData` was not intended to be called often - it's more a convenience function to extract all the data as a `DataFrame`. In day-to-day use users should use the `Spectra` object itself - it can be used like a `DataFrame` anyway, i.e. it is possible to access any spectra variable with `$`.

lgatto:
[...]

PR #94, opened 2020-03-05
refactor: mz and intensity aren't part of the default spectraVariables anymore

lgatto:
This PR addresses #92. The `mz` and `intensity` variables aren't part of the default `spectraVariables` any more and hence won't populate the `spectraData`, making it slow to access and big.

This PR only affects the on disk backends and not the in memory DataFrame backend, because for the latter, `mz` and `intensity` are already part of the DataFrame (and this aren't added).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: Spectra
 Title: Spectra Infrastructure for Mass Spectrometry Data
-Version: 0.4.1
+Version: 0.4.2
 Description: The Spectra package defines a efficient infrastructure
    for storing and handling for raw mass spectrometry spectra.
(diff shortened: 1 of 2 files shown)


lgatto:
Sorry, some unit tests are failing here. Will update this PR asap.

lgatto:
One side effect of this change is that this (taken from `testthat.R`) now fails - `sciex_hd5` contains no `mz` or `intensity` values. 

```r
sciex_hd5 <- backendInitialize(MsBackendHdf5Peaks(),
                               spectraData = spectraData(sciex_mzr),
                               hdf5path = tempdir())
```

Easy fix is to add manually, which I'm going to do for now. 

But @jorainer, do you know if that pattern is used widely?

PR #95, opened 2020-03-06
spectraData,Spectra does not return m/z and intensity values

jorainer:
This PR (discussed in #92)
- replaces the `spectraData,MsBackend` method with `asDataFrame,MsBackend` (and `spectraData<-,MsBackend` with `asDataFrame<-,MsBackend`.
- changes `spectraVariables,Spectra` to **not** return `"mz"` and `"intensity"` with the result that `spectraData,Spectra` does return only a `DataFrame` with spectra metadata, no m/z and intensity values.
- add a method `dropNaSpectraVariables` to allow application of this function to both `Spectra` and `MsBackend` (the function that was in place before used `spectraData`, that now however does no longer exist for `MsBackend`. I thought this function might still be useful for `MsBackend` and that's why I added it as a method.
- Updates the documentation and fixes the unit tests.


jorainer:
@sgibb, should I wait for your review or am I OK to merge?

R/Spectra.R
```diff
@@ -989,6 +987,15 @@ setReplaceMethod("dataOrigin", "Spectra", function(object, value) {
 #' @rdname Spectra
 setMethod("dataStorage", "Spectra", function(object) dataStorage(object@backend))
 
+#' @rdname Spectra
+setMethod("dropNaSpectraVariables", "Spectra", function(object) {
```
sgibb: This method is nearly identical to `dropNaSpectraVariables,MsBackend`. Wouldn't be the following sufficient:
```r
setMethod("dropNaSpectraVariables", "Spectra", function(object) {
    object@backend <- dropNaSpectraVariables(object@backend)
    object
}
```

Another solution would be to drop  `dropNaSpectraVariables,MsBackend` (and keep  `dropNaSpectraVariables,Spectra`) because it isn't used anywhere, or?

jorainer: thanks for the suggestion. The `dropNaSpectraVariables,MsBackend` isn't used anywhere yet, but I find it quite usefull and we might use/call it also in the backend at some point.

jorainer: Done.

PR #97, opened 2020-05-14
refactor: import ndotproduct instead of dotproduct from MsCoreUtils

jorainer:
Import `ndotproduct` instead of `dotproduct` from `MsCoreUtils`.


jorainer:
I'm going to boldly merging this PR without review - as it just imports the `ndotproduct` from `MsCoreUtils` instead of the `dotproduct`.

Issue #82, opened 2019-12-11
Error reading some data files


jorainer:
The following code creates an error

```r
> fls <- dir("/Users/jo/data/mzML/2019/2019_10", full.names = TRUE)
> sps <- Spectra(fls, source = MsBackendMzR(), backend = MsBackendDataFrame())
Error in map_x_colnames_to_object_colnames(colnames(object)) : 
  the DataFrame objects to rbind must have the same number of columns
```

jorainer:
This error was caused by different set of spectra variables returned (i.e. some of the files have and some don't have MS2 spectra).

PR #99, opened 2020-06-24
fix: support import of mzML files with different header columns

jorainer:
- This fixes issue #82 by using `rbindFill` instead of `rbind` in combining the
  header of the individual mzML files into the final `DataFrame`.
- Note that this will also require `MsCoreUtils` 1.1.4 that fixes an issue in `rbindFill` if only a single `DataFrame` is passed.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: Spectra
 Title: Spectra Infrastructure for Mass Spectrometry Data
-Version: 0.5.2
+Version: 0.5.3
 Description: The Spectra package defines a efficient infrastructure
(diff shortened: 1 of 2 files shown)


jorainer:
Merging it because it's really just the replacement of `rbind` with `rbindFill`

DESCRIPTION
```diff
@@ -25,7 +25,7 @@ Depends:
 Imports:
     methods,
     IRanges,
-    MsCoreUtils (>= 1.1.1),
+    MsCoreUtils (>= 1.1.4),
```
sgibb, line 28: I just pushed `MsCoreUtils` 1.1.**3** to bioc. I decrement the version number in master already.

Issue #105, opened 2020-07-03
CI: travis or GHA


lgatto:
We currently use travis and GitHub actions. I suggest we drop the former. 

@sgibb @jorainer

jorainer:
Absolutely. And we then also enable the packagedown step in the github action.

lgatto:
Ah, that isn't in yet (good I asked before breaking it). Feel free to send a PR, otherwise I will look into another repo to learn how to do it.

jorainer:
Also pkgdown works with GA. Closing issue.

lgatto:
I have disabled `Spectra` on travis - it seemed to still be build despite the lack of config file.

PR #107, opened 2020-07-03
ci: remove Travis and enable packagedown in GA (issue #105)

jorainer:
Let's see if this works out-of-the-box


jorainer:
I will merge this to check if `pkgdown` is properly running.

Issue #80, opened 2019-12-11
Add some basic plotting functions


sgibb:
In the "ms proteomics" user/developer session we discussed some `plot` functions.

We want to:
- use only *base graphics* (no `ggplot2`, no `lattice`, ...).
- use unit tests with `vdiffr`.
- keep it simple and basic (no annotation, no ion mobility).

Initial planned functions:
- `plot(Spectra[1])` plots a single spectrum (as `type = "l"` for profile (default) and `type = "h" for centroided).
- `plot(Spectra[1:10])` plots multiple spectra above each other as stacked plot or as *messy*/coplot plot all in the same plot area.
- `lcmsMap(Spectra)`: "3D" plot, rt on the x-axis, mz on the y-axis, intensity colour-coded.

(hand-crafted image is following)

![IMG_1566](https://user-images.githubusercontent.com/5506112/70623715-fdbf7100-1c1e-11ea-8dc9-9009a68323de.JPG)

See example stub code:
https://github.com/rformassspectrometry/Spectra/blob/bca663197cb18471d92613a9b4417d9b9f3adeb8/R/plot-functions.R#L1-L26

CC: @cpanse, @tobiasko

Please have a look at the [plot branch](https://github.com/rformassspectrometry/Spectra/tree/plot).

tobiasko:
Cool! Regrading unit testing. I guess you already saw this @sgibb , but just in case...Accoiding to [stackoverflow](https://stackoverflow.com/a/44990728) [vdiffr](https://github.com/r-lib/vdiffr) is used by gglot for unit testing.

sgibb:
Thanks @tobiasko, I already add a small unit test for the very basic `.plotSingleSpectrum` function:
https://github.com/rformassspectrometry/Spectra/blob/ad405ee94f59e02c9ac7b0582c8dd7ef7b29cbe5/tests/testthat/test_plot-functions.R#L13-L25

tobiasko:
Maybe a stupid question, but why do `Spectra::mz(...)` and `Spectra::intensity(...)` return a list instead of a vector, also for `length(Spectra) == 1` ? Connected, my naive  `.plotSingleSpectrum(...)` would use the getter functions  also internally:
```
## Create a Spectra providing a `DataFrame` containing the spectrum data.
spd <- DataFrame(msLevel = c(1L, 2L), rtime = c(1.1, 1.2))
[...]

PR #108, opened 2020-07-03
Add plotSpectraOverlay (issue #80)

jorainer:
- Add the `plotSpectraOverlay` function that allows to plot all spectra into the same plot.
- Add a section describing the plotting functions to the vignette.


lgatto:
Thank you!

PR #115, opened 2020-08-03
Improve .peaksapply performance

jorainer:
Improve performance of the `.peaksapply` call avoiding `split`, `bplapply` and `unsplit` where not necessary. This results in a 2x performance gain.


jorainer:
This PR includes also a fix to the plotting unit tests.

R/Spectra-functions.R
```diff
@@ -80,7 +80,7 @@ addProcessing <- function(object, FUN, ...) {
 #'
 #' @noRd
 .peaksapply <- function(object, FUN = NULL, ..., f = dataStorage(object),
-                        BPPARAM = bpparam()) {
+                         BPPARAM = bpparam()) {
```
sgibb, line 83: Is this space intended?
```suggestion
                        BPPARAM = bpparam()) {
```

Issue #116, opened 2020-08-05
Improve performance of compareSpectra


jorainer:
The `compareSpectra` function is not as fast as I would like to have it. Seems that the iterative calling of the `.peaksapply` function to *realize* the peak matrix (one at a time) is the main bottleneck. It would be faster to *realize* all peak matrices first and then to loop over them:

```r
.compare_spectra_mem <- function(x, y, MAPFUN = joinPeaks, tolerance = 0,
                                 ppm = 20, FUN = ndotproduct, ...) {
    mat <- matrix(NA_real_, nrow = length(x), ncol = length(y),
                  dimnames = list(spectraNames(x), spectraNames(y)))
    x <- Spectra:::.peaksapply(x)
    y <- Spectra:::.peaksapply(y)
    for (i in seq_along(x)) {
        for (j in seq_along(y)) {
            peak_map <- MAPFUN(x[[i]], y[[j]],
                               tolerance = tolerance, ppm = ppm, ...)
            mat[i, j] <- FUN(peak_map[[1L]], peak_map[[2L]], ...)
            }
    }
    mat
}

library(microbenchmark)
microbenchmark(
    Spectra:::.compare_spectra(all_sps[1], all_sps[1:1000]),
    .compare_spectra_mem(all_sps[1], all_sps[1:1000]),
    times = 5)

Unit: milliseconds
                                                    expr      min        lq
 Spectra:::.compare_spectra(all_sps[1], all_sps[1:1000]) 5478.865 5814.8966
       .compare_spectra_mem(all_sps[1], all_sps[1:1000])  183.608  188.5381
      mean    median        uq       max neval cld
 5869.4490 5880.5818 6046.3839 6126.5182     5   b
  190.5342  190.2187  190.8331  199.4731     5  a 
```

However, requiring all data in memory would not work for very large scale data sets. One possibility would be to process the data in chunks, i.e. define a chunk size, e.g. 1000 and then always *realize* this number of spectra in memory and loop over them.

jorainer:
I've implemented a `compare_spectra_chunk` function that realizes peak data in chunks. The performance comparison is shown below:

```r
library(Spectra)
sciex_file <- normalizePath(
    dir(system.file("sciex", package = "msdata"), full.names = TRUE))
sciex_mzr <- backendInitialize(MsBackendMzR(), files = sciex_file)

sps_mzr <- Spectra(sciex_mzr)
[...]

PR #118, opened 2020-08-05
Add chunk-based data loading in compareSpectra

jorainer:
This PR adds the chunk-based data loading for `compareSpectra` as discussed in #116 .


jorainer:
Uh - just realized that this PR contains also PR #115

R/Spectra-functions.R
```diff
@@ -240,50 +248,42 @@ applyProcessing <- function(object, f = dataStorage(object),
...
-            for (i in x_idx) {
-                peak_map <- MAPFUN(.peaksapply(x[i])[[1]], py,
-                                   tolerance = tolerance, ppm = ppm, ...)
-                mat[i, j] <- FUN(peak_map[[1L]], peak_map[[2L]],
-                                 ...)
-            }
+    for (x_chunk in x_chunks) {
+        for (y_chunk in y_chunks) {
+            mat[x_chunk, y_chunk] <- .peaks_compare(
+                .peaksapply(x[x_chunk]), .peaksapply(y[y_chunk]),
```
sgibb, line 284: Wouldn't it be faster to put the `.peaksapply(x[x_chunk])` into the outer `for` loop instead of the inner (as in the former implementation)? Currently it is called `length(x_chunks) * length(y_chunks)` times instead of just `length(x_chunks)` times.

R/Spectra-functions.R
```diff
@@ -240,50 +248,42 @@ applyProcessing <- function(object, f = dataStorage(object),
...
+    }
 
-    nx <- length(x_idx)
-    ny <- length(y_idx)
+    x_idx <- seq_len(nx)
+    x_chunks <- split(x_idx, ceiling(x_idx / chunkSize))
+
+    y_idx <- seq_len(ny)
+    y_chunks <- split(y_idx, ceiling(y_idx / chunkSize))
+    rm(x_idx)
```
sgibb, line 276: Just minor coding style point: I would put the `rm(x_idx)` direct below `x_chunks <- split(x_idx, ceiling(x_idx / chunkSize))`. BTW do we need the chunk calculation in other functions as well? If that is the case I would put it in a separate function. Maybe in `MsCoreUtils` if you need this in `Chromatograms` etc.

R/Spectra-functions.R
jorainer: Makes total sense! Thanks for spotting (another little performance gain 👍 )!

R/Spectra-functions.R
jorainer: OK, I will move the `rm(x_idx)`.

Since it is a one-liner I would keep the chunk calculation for now within this function. Maybe later - if we indeed will need it also for `Chromatograms` we could define a function for it.

sgibb (approved):
Fine. Go ahead!

Issue #124, opened 2020-08-27
New export method


lgatto:
- Serialise any `Spectra` object to mgf.
- Also (later) serialise to mzML.

jorainer:
We may rethink the concept of exporting the data eventually by adding an `MsBackend,export` method. Tagging @michaelwitting @meowcat @sgibb and @lgatto

jorainer:
Note that a simple dispatching of the `export,Spectra` to `export,MsBackend` will **not** work, because the backend does not know anything about data manipulations that have been happening in the `Spectra` object (i.e. that are in the processing queue).

jorainer:
So, IMHO the best would be to have a `export,MsBackend` method that takes a `Spectra` object as input. The definition of the `export` could be:

```r
setMethod("export", "Spectra", "MsBackend", function(object, backend, file, ...) {
    ## use here the `spectraData` method to get all metadata of the Spectra
    ## and the `as.list` to get the peaks (m/z and intensity values)
})
```

and this function should be defined in the package providing the backend (e.g. `MsBackendMgf`). I will tinker a little with this in the `MsBackenMgf` package to see if that works or if there is some unforseen problem with it.

lgatto:
I thought that `export,MsBackend` had always been on the table (but maybe I'm confused with `reset`).

Why could `export,Spectra` not call `export,MsBackend` **after** somehow handling the delayed processing/filtering operations? I am not keen on `export,MsBackend` hanlding a `Spectra` object. I think the devil is in the details here, in particular on the type of  on-disk backend.

jorainer:
I know, it doesn't sound intuitive that the backend should deal with a `Spectra`. 

Your suggestion with handling the processing queue would however mean to use `applyProcessing` on the `Spectra`, which replaces peak values in the backend, something that read-only backends, such as the `MsBackendMzR` would not support. The alternative would be to call `.peaksapply` in the `export,Spectra` and pass then this `list` of peak matrices to the `export,MsBackend` - so we would have in the worst case have twice the peak data in memory - I would like to avoid that, also to support large scale mzML-based experiments. 

IMHO it would be an easier and cleaner implementation if we passed the `Spectra` object to the backend - in the simple case the backend export method has to simply call the `spectraData` and `as.list` methods on the `Spectra`, but we would also allow for other implementations, such as for `MsBackendMzR` where I would like to perform the export on a per-file basis and would only load the data from one file at a time to be more memory efficient.

[...]

PR #125, opened 2020-08-27
feat: add export,Spectra method (issue #124)

jorainer:
- Add `export,Spectra` to export spectra data to a file in MGF format.
- Add related unit tests and documentation.


R/Spectra.R
```diff
@@ -83,6 +83,31 @@ NULL
...
 #'
+#' Data from a `Spectra` object can be **exported** to a file with the `export`
+#' function. The file format for the exported data can be specified with the
+#' parameter `format`; currently only `format = "mgf"` is supported which
+#' exports the data in Mascot Generic Format (MGF). By default all non-empty
+#' spectra variables in `Spectra` are exported.
+#'
+#' The definition of the function is
+#' `export(object, file = tempfile(), format = "mgf",
+#'     variableMapping = spectraVariableMapping(format), ...)` and ist
```
lgatto, line 94: and *its*

R/Spectra-functions.R
```diff
@@ -484,3 +484,66 @@ combineSpectra <- function(x, f = x$dataStorage, p = x$dataStorage,
         else common(y, z, tolerance = tolerance, ppm = ppm)
     })
 }
+
+#' @export
+#'
+#' @rdname Spectra
+spectraVariableMapping <- function(format = c("mgf")) {
```
lgatto, line 491: Isn't there something like this is `MsBackendMgf`? Anyway, that mapping should be defined only once, and `Spectra` is the only package for that. it will this be important (note to self) to reuse it in `MsBackendMgf`.

lgatto (changes requested):

R/Spectra-functions.R
jorainer: Agree. At first glance I haven't seen anything in `MsBackendMgf`.

R/Spectra.R
jorainer: done in https://github.com/rformassspectrometry/Spectra/pull/125/commits/4be7b958b75c680ae660575ac1a53c81b7d7dfa7

lgatto (approved):

Issue #133, opened 2020-09-14
Filtering spectra by m/z


jorainer:
There are two different use cases of filtering a spectrum by m/z:
- filter by an m/z range (that's what the `MSnbase::filterMz` is doing): keep only peaks that are within the provided m/z range.
- filter a spectrum keeping only peaks that match the provided m/z value(s). This is something I need at present: given a spectrum, find/filter to the peaks that match a set of m/z values (from different ions of a compound).

We could now either:
1) implement two separate filter methods `filterMzRange` for the first and `filterMz` for the latter use case.
2) implement a single `filterMz` with a parameter `range = TRUE` to switch between range- or match-based filtering.

@sgibb @lgatto any comments, ideas?

lgatto:
I prefer two separate functions rather than a parameter and would suggest
- `filterMzRange()`
- `filterMzValues()`

jorainer:
Fine for me. Thanks @lgatto

PR #134, opened 2020-09-14
feat: add filterMzRange and filterMzValues

jorainer:
- Add `filterMzRange,Spectra` and `filterMzValues,Spectra` to subset spectra
  keeping only peaks that have either an m/z within the specified range or peaks
  with m/z values matching the provided m/z value(s) (issue #133).
- Add related documentation and unit tests.


jorainer:
Seems I get an unexpected error in an unit test that I don't get locally. I will investigate.

jorainer:
Problem is that `Spectra` now needs a more recent `MsCoreUtils` package than the version 1.1.3 which is in BioC. Also, it seems that the package builder skips the `MsCoreUtils` package as it is still at version 1.1.3 although in the Bioc git it is already at 1.1.5. I send a mail to BioC-devel to report this.

jorainer:
OK, checks run now. So green lights for the review :)

R/peaks-functions.R
```diff
@@ -90,6 +90,54 @@ NULL
...
+#'
+#' @noRd
+.peaks_filter_mz_value <- function(x, spectrumMsLevel, mz = numeric(),
+                                   tolerance = 0, ppm = 20,
+                                   msLevel = spectrumMsLevel, ...) {
+    if (!spectrumMsLevel %in% msLevel || !length(x))
+        return(x)
+    idx <- closest(mz, x[, "mz"], tolerance = tolerance, ppm = ppm,
+                   duplicates = "closest")
+    x[idx[!is.na(idx)], , drop = FALSE]
```
sgibb, line 119: That's exactly what `common` does:
```suggestion
    keep <- common(mz, x[, "mz"], tolerance = tolerance, ppm = ppm,
                   duplicates = "closest")
    x[keep, , drop = FALSE]
```

R/peaks-functions.R
```diff
@@ -90,6 +90,54 @@ NULL
...
+#' @param mz `numeric(2)` with the lower and upper m/z value.
+#'
+#' @author Johannes Rainer
+#'
+#' @noRd
+.peaks_filter_mz_range <- function(x, spectrumMsLevel, mz = numeric(),
+                                   msLevel = spectrumMsLevel, ...) {
+    if (!spectrumMsLevel %in% msLevel || !length(x))
+        return(x)
+    x[which(between(x[, "mz"], mz)), , drop = FALSE]
```
sgibb, line 138: Subsetting with `which` was faster in earlier days but now it just offers a benefit if we subset multiple times using the same index:
```r
set.seed(123)
x <- sample(10, 1e2, TRUE)
microbenchmark::microbenchmark(lgl=x[x > 5], which=x[which(x > 5)])
# Unit: microseconds
#   expr   min    lq    mean median     uq   max neval cld
#    lgl 1.098 1.264 1.50135 1.4205 1.6000 7.629   100  a 
#  which 1.586 2.052 2.46258 2.2380 2.4595 9.091   100   b

x <- sample(10, 1e6, TRUE)
microbenchmark::microbenchmark(lgl=x[x > 5], which=x[which(x > 5)])
# Unit: milliseconds
#   expr      min       lq      mean   median       uq      max neval cld
#    lgl 8.180921 8.447758  9.985675 9.222107 11.01384 17.56540   100   a
#  which 8.958257 9.192004 12.349117 9.915709 11.43943 98.04092   100   a
```
IMHO there was a NEWS entry for R 3.x.x somewhere but I can't find it anymore.

sgibb (changes requested):
Thanks for this PR. Just a few small things to discuss.

R/Spectra.R
```diff
@@ -1508,6 +1554,48 @@ setMethod("filterMsLevel", "Spectra", function(object, msLevel. = integer()) {
...
+setMethod("filterMzRange", "Spectra",
+          function(object, mz = numeric(), msLevel. = unique(msLevel(object))) {
+              if (!.check_ms_level(object, msLevel.))
+                  return(object)
+              mz <- range(mz)
+              object <- addProcessing(object, .peaks_filter_mz_range,
+                                      mz = mz, msLevel = msLevel.)
+              object@processing <- .logging(
+                  object@processing, "Filter: select peaks with an m/z within ",
+                  "[", paste0(mz, collapse = ", "), "]")
```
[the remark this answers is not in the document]
jorainer: You're right. I will change that ;)

R/Spectra.R
```diff
@@ -1508,6 +1554,48 @@ setMethod("filterMsLevel", "Spectra", function(object, msLevel. = integer()) {
...
+setMethod("filterMzValues", "Spectra",
+          function(object, mz = numeric(), tolerance = 0, ppm = 20,
+                   msLevel. = unique(msLevel(object))) {
+              if (!.check_ms_level(object, msLevel.))
+                  return(object)
+              if (is.unsorted(mz)) {
+                  idx <- order(mz)
+                  mz <- mz[idx]
+                  l <- length(mz)
+                  if (length(tolerance) == l)
```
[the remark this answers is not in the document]
jorainer: Actually, I thought that this would be catched downstream by the `closest` call - but maybe you're right and it might be safer (and also immediate) to test it here.

R/Spectra.R
```diff
@@ -1508,6 +1554,48 @@ setMethod("filterMsLevel", "Spectra", function(object, msLevel. = integer()) {
...
+                      tolerance <- tolerance[idx]
+                  if (length(ppm) == l)
+                      ppm <- ppm[idx]
+              }
+              object <- addProcessing(object, .peaks_filter_mz_value,
+                                      mz = mz, tolerance = tolerance,
+                                      ppm = ppm, msLevel = msLevel.)
+              object@processing <- .logging(
+                  object@processing, "Filter: select peaks matching provided ",
```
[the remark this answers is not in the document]
jorainer: Sure, I can change that.

R/peaks-functions.R
jorainer: My main reason to use `which` is to drop any `NA`s. If we can be sure that `between` never returns `NA` we can change - otherwise I would keep it to be on the save side because I think we have to avoid having peaks with `NA` m/z values.

R/Spectra.R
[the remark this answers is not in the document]
jorainer: Done in https://github.com/rformassspectrometry/Spectra/commit/927c6ad32f4ce3fa2330713c1774c4874827541c

R/Spectra.R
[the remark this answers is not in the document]
jorainer: Done in https://github.com/rformassspectrometry/Spectra/commit/927c6ad32f4ce3fa2330713c1774c4874827541c

R/Spectra.R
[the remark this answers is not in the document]
jorainer: Done in https://github.com/rformassspectrometry/Spectra/commit/927c6ad32f4ce3fa2330713c1774c4874827541c

R/peaks-functions.R
jorainer: Done in https://github.com/rformassspectrometry/Spectra/commit/a80db29a4601e43b176513e4a364f75e1af14a23

R/peaks-functions.R
jorainer: OK, so `between` throws an error if there are `NA`s, so it's safe to drop the `which` here. Done in https://github.com/rformassspectrometry/Spectra/commit/02997e36be9e8b53d4937d53b2788f3b3e24c5da

R/Spectra.R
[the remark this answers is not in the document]
sgibb: Oh, yes you are right. Duplicating this check seems to be useless. So may we should use the `.check` argument in `closest` to avoid that check as well? (because here it is run once and in `closest` `length(object)` times)

sgibb (approved):
All good from my side!

R/peaks-functions.R
jorainer: Actually, I had to revert this. Calling `common` in that way returns a `logical` of length 1. If we call it with `common(x[, "mz"], mz, ...)` it would return a logical with the correct length, but then `tolerance` and `ppm` can no longer be of length `= length(mz)`.

jorainer:
I agree. I call now `closest` with `.check = FALSE`. I'm just waiting for the local tests to pass before pushing.

R/peaks-functions.R
jorainer: Would like to get your OK on reverting that back @sgibb

tests/testthat/test_peaks-functions.R
```diff
@@ -167,14 +167,16 @@ test_that(".peaks_filter_mz_range works", {
...
+    expect_true(nrow(res) == 0)
     expect_equal(colnames(res), c("mz", "intensity"))
-    res <- .peaks_filter_mz_value(p, 1L, mz = NA)
+    res <- Spectra:::.peaks_filter_mz_value(p, 1L, mz = NA)
     expect_true(is.matrix(res))
     expect_equal(colnames(res), c("mz", "intensity"))
+    expect_true(nrow(res) == 0)
 
-    res <- .peaks_filter_mz_value(p, 1L, mz = 5, tolerance = 1)
+    res <- Spectra:::.peaks_filter_mz_value(p, 1L, mz = 5, tolerance = 1)
```
sgibb, line 179: Is that `Spectra:::` needed? If yes it is missing two lines below as well.

sgibb (approved):
Somehow I can't answer directly under the corresponding code anymore.

You are right. I overlooked the problem with `common`. That's just happened because you wanted me to change `tolerance = length(table)` to `tolerance = length(x)` in `closest` :smile:

tests/testthat/test_peaks-functions.R
jorainer: That's not required. I just forgot to remove them.

jorainer: Will remove.

jorainer:
Thanks for your review @sgibb !

Issue #138, opened 2020-09-24
dropNaSpectraVariables drop m/z and intensity


michaelwitting:
Dear RforMassSpectrometry-Team,

when reading MS2 spectra from .mgf or Massbank files the `dropNaSpectraVariables` removes values in `mz` and `intensity`.
Any ideas?

Best regards,

Michael

Example file attached

```
ms2_spectra <- Spectra("prx-5_1_1-A,4_01_35913.mgf",
                                       source = MsBackendMgf(),
                                       backend = MsBackendDataFrame())

mz(dropNaSpectraVariables(ms2_spectra))
```

[prx-5_1_1-A,4_01_35913.zip](https://github.com/rformassspectrometry/Spectra/files/5274492/prx-5_1_1-A.4_01_35913.zip)

sgibb:
@michaelwitting thanks for reporting this bug!

`MsBackendMgf` inherits this method from `MsBackend`. A MWE:

```r
d <- DataFrame(msLevel = 1L, scanIndex = 1L)
d$mz <- list(c(1:3))
d$intensity <- list(c(4:6))

r <- Spectra(d)
mz(dropNaSpectraVariables(r))
```

sgibb:
`spectraVariables,Spectra` removes `mz` and `intensity`:

https://github.com/rformassspectrometry/Spectra/blob/0c505e159a72ce030fc4b86f9dc05e93340d6404/R/Spectra.R#L1431-L1434

```r
spectraVariables(r@backend)
#  [1] "msLevel"                 "rtime"
#  [3] "acquisitionNum"          "scanIndex"
#  [5] "mz"                      "intensity"
#  [7] "dataStorage"             "dataOrigin"
#  [9] "centroided"              "smoothed"
# [11] "polarity"                "precScanNum"
# [13] "precursorMz"             "precursorIntensity"
# [15] "precursorCharge"         "collisionEnergy"
# [17] "isolationWindowLowerMz"  "isolationWindowTargetMz"
# [19] "isolationWindowUpperMz"
spectraVariables(r)
#  [1] "msLevel"                 "rtime"
#  [3] "acquisitionNum"          "scanIndex"
#  [5] "dataStorage"             "dataOrigin"
#  [7] "centroided"              "smoothed"
#  [9] "polarity"                "precScanNum"
# [11] "precursorMz"             "precursorIntensity"
# [13] "precursorCharge"         "collisionEnergy"
[...]

PR #139, opened 2020-09-30
fix: dropNaSpectraVariables no longer drops mz and intensity

jorainer:
- Fix `dropNaSpectraVariables` does no longer drop the `"mz"` and `"intensity"`
  columns (issue #138).
- Update also documentation of `spectraData,MsBackend` to clearly state it has
  to return also m/z and intensity values.


sgibb (approved):
:tada: 2nd approved PR without comments :smile:

Issue #143, opened 2020-10-15
Backend confusion


ococrook:
Hi - me again...

I'm confused by what "backend = " is doing.

The following code chunk doesn't seem to do what I expected. Specifying 'backend = MsBackendDataFrame()' seems to set MsBackendMzR. Though it happily changes the backend using setBackend.

```{r,}
fl <- "data/191115_PhosB_HDMSe_0sec_01_vial01.mzML"
PhosB_raw <- Spectra(fl, backend = MsBackendDataFrame())
PhosB_raw
PhosB_raw <- setBackend(PhosB_raw, MsBackendDataFrame())
PhosB_raw
```
MSn data (Spectra) with 362200 spectra in a MsBackendMzR backend:
         msLevel     rtime scanIndex
       <integer> <numeric> <integer>
1              1     1.422         1
2              1     1.422         2
3              1     1.422         3
4              1     1.422         4
5              1     1.422         5
...          ...       ...       ...
362196         2   659.956    362196
362197         2   659.956    362197
362198         2   659.956    362198
362199         2   659.956    362199
362200         2   659.956    362200
 ... 33 more variables/columns.

file(s):
191115_PhosB_HDMSe_0sec_01_vial01.mzML
MSn data (Spectra) with 362200 spectra in a MsBackendDataFrame backend:
         msLevel     rtime scanIndex
       <integer> <numeric> <integer>
1              1     1.422         1
2              1     1.422         2
3              1     1.422         3
4              1     1.422         4
5              1     1.422         5
...          ...       ...       ...
362196         2   659.956    362196
362197         2   659.956    362197
362198         2   659.956    362198
362199         2   659.956    362199
362200         2   659.956    362200
 ... 33 more variables/columns.
Processing:
 Switch backend from MsBackendMzR to MsBackendDataFrame [Thu Oct 15 11:28:18 2020]

jorainer:
What happens when you use

```r
PhosB_raw <- Spectra(fl, source = MsBackendMzR(), backend = MsBackendDataFrame())
```

With `source` you would then specify that you use the mzR backend for import and then it *should* change (after import) to the `MsBackendDataFrame()`. I know it's a little confusing...

ococrook:
It looks the same to me

```
> fl <- "data/191115_PhosB_HDMSe_0sec_01_vial01.mzML"
> PhosB_raw <- Spectra(fl, source = MsBackendMzR(), backend = MsBackendDataFrame())
> PhosB_raw
[...]

PR #144, opened 2020-10-15
fix: ensure backend is changed in Spectra,character constructor

jorainer:
- Fix that the backend is changed even if the backend passed with `source`
  extends the backend specified with `backend` (issue #146).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: Spectra
 Title: Spectra Infrastructure for Mass Spectrometry Data
-Version: 0.99.8
+Version: 0.99.9
 Description: The Spectra package defines an efficient infrastructure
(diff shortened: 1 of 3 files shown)


lgatto:
Looks good, but could you point me to the issue or example that triggered this change. I can only vaguely remember it.

jorainer:
Sure, sorry @lgatto , I linked a wrong issue. The correct one is #143 .

lgatto (approved):

lgatto:
Merged now. Olly, could you install from Github with 
```r
BiocManager::install("RforMassSpectrometry/Spectra")
```
and confirm that it now works.

lgatto:
@jorainer - I tried to push to Bionconductor, but don't have access. Will email Nitesh and cc you.

lgatto:
On Bioc (devel) too, now.

Issue #146, opened 2020-10-19
manipulating spectraData clears peaksData


stanstrup:
```r
spd <- DataFrame(msLevel = c(1L, 2L), rtime = c(1.1, 1.2))
spd$mz <- list(c(100, 103.2, 104.3, 106.5), c(45.6, 120.4, 190.2))
spd$intensity <- list(c(200, 400, 34.2, 17), c(12.3, 15.2, 6.8))

data <- Spectra(spd)

peaksData(data) %>% as.list()
```

```r
[[1]]
        mz intensity
[1,] 100.0     200.0
[2,] 103.2     400.0
[3,] 104.3      34.2
[4,] 106.5      17.0

[[2]]
        mz intensity
[1,]  45.6      12.3
[2,] 120.4      15.2
[3,] 190.2       6.8
```

Now changing spectraData and peaksData becomes empty:
```r
spectraData(data) <- spectraData(data)
peaksData(data) %>% as.list()
```

```r
[[1]]
     mz intensity

[[2]]
     mz intensity
```

I assume this is not intended?

jorainer:
That's a tricky one. Yes, I agree that it is not intuitive that this function removes m/z and intensity values since `spectraData` does not return these values. I will fix that by internally re-adding the `mz` and `intensity` columns if they are not present in `value`. This can however cause problems if the new spectra data is in a different order than before!

Thus, if possible, I suggest replacing individual spectra variables/columns instead of replacing the full data at once. I will update the documentation accordingly.

stanstrup:
OK. Thanks.
So this should be the way to go?:

```r
data@backend@spectraData$msLevel <- c(3,4)
```

jorainer:
No, please don't access the backend directly. It would be `msLevel(data) <- c(3, 4)` or `data$msLevel <- c(3, 4)`. All operations should be possible on the `Spectra` object, thus the *normal* users should never access the backend.

jorainer:
Anyway, what you reported is a bug, because `spectraData <- ` on a `Spectra` object should not remove m/z and intensity values.

stanstrup:
Aha. OK. I forgot about directly access the `spectraData` with `$`.

jorainer:
And you can also add additional spectra variables with that, e.g. `data$new_col <- "a"` would add a new spectra variable `"new_col"`.

jorainer:
It's fixed with the new version. Closing the issue now - feel free to reopen if needed @stanstrup

PR #147, opened 2020-10-19
fix: replacing spectraData on Spectra does not remove m/z

jorainer:
- Avoid removing m/z and intensity values when calling
  `spectraData<-,Spectra` (issue #146).


jorainer:
An additional change in this PR is that I moved most of the `filter*` functions from `MsBackendDataFrame` to `MsBackend` hence avoiding the need to re-implement these methods for all new backends.

lgatto (approved):

jorainer:
@lgatto , did you already push to Bioconductor?

lgatto:
Was about to, but it looks like you have just done it.

jorainer:
But there are some strange things - like the github and bioconductor masters are not in sync - please see PR #148 (please don't merge that one!).

PR #148, opened 2020-10-20
Master upstream

jorainer:


R/MsBackendDataFrame-functions.R
```diff
@@ -85,8 +85,6 @@ NULL
     rtime = "numeric",
     acquisitionNum = "integer",
     scanIndex = "integer",
-    mz = "NumericList",
```
jorainer, line 88: @lgatto what was the reason to remove `mz` and `intensity` here? I think that removing these lines will cause some downstream problems. Also, what puzzled me is that this seems to be present in the Bioconductor master branch but not (yet) in the github branch?

jorainer:
I had a conflict between my local master and the Bioconductor master and was wondering why - seems that these changes here are in Bioconductor master, but not in our github master. I think we should revise this and check if these changes are really necessary - and in addition we need to sync the github and Bioconductor master

lgatto:
I don't remember of these lines were to be removed, but this was part of the discussion whether `mz` and `intensity` where part of the spectra variables and whether they should be returned as part of `spectraData()`.

Weird re out of sync of Github and Bioc - this should indeed never happen :-/

jorainer:
In fact `mz` and `intensity` should **not** be removed from the `.SPECTRA_DATA_COLUMNS`, these are used in the validation functions of the `MsBackendDataFrame` object. Also, `mz` and `intensity` have to be part of the `spectraData` `DataFrame` returned from a `MsBackend` - always, because that is used to transfer the data if the backend is changed.

The `Spectra` object should by default not report these columns. But that's in the `Spectra` object, not in the `MsBackend`.

What I will do is
a) re-add `mz` and `intensity`
b) push these changes also to Bioconductor and ensure that everything is in sync again.

lgatto:
Not sure why they were removed to be honest. Sorry if I did that, must have happened by mistake.

jorainer:
No worries - I re-added them. I will then push also to Bioc once all tests passed

lgatto:
Thanks!

PR #153, opened 2020-10-28
Round retention time in plot titles

lgatto:


lgatto:
In addition to the current PR, I would suggest to also mention the MS level in the title. Something like

```
MS1 - RT 1800.7
```

lgatto:
The failing actions are a matter of wrong R version (4.1 required) on Windows and OSX. The tests have been updated to account for the change in the plotting functions.

jorainer (approved):
All OK. Thanks for the PR @lgatto - feel free to merge.

Issue #152, opened 2020-10-28
Add a new Spectra constructor for DBIConnection


jorainer:
Maybe a `Spectra` constructor for `object = "DBIConnection"` might be helpful?

```r
setMethod("Spectra", "DBIConnection", function...) {
  be <- backendInitialize(backend, object, ...)
  ...
}
```
i.e. that takes a database connection as input and calls the `backendInitialize` method on the provided backend passing that argument as parameter to it.

Would this also work for your backend @plantton?

jorainer:
Actually, thinking it over I don't like the idea anymore - this would add `DBI` as a required R package, while the `Spectra` package itself does not contain any SQL-related functionality. So, no I don't want that constructor to be present in `Spectra`.

PR #156, opened 2020-10-29
refactor: replace Spectra,DataFrame by Spectra,ANY

jorainer:
Implement a `Spectra,ANY` constructor method that *imports* data using the `source` backend (eventually converting to `backend` backend). This supports also to create a `Spectra` object based on a e.g. database connection, given that the `MsBackend` specified with `source` knows how to handle that (issue #152).

I think this will save us from unnecessary additional implementation of constructor methods, e.g. for `Spectra,DBIconnection` or similar.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: Spectra
 Title: Spectra Infrastructure for Mass Spectrometry Data
-Version: 1.1.2
+Version: 1.1.3
 Description: The Spectra package defines an efficient infrastructure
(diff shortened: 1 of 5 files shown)


lgatto:
Very elegant, thanks!

lgatto (approved):

jorainer:
If OK I would also make this change in the release version of the package. This shouldn't break anything but allow us to directly use this new constructor...

lgatto:
Yes, please go ahead.

Issue #149, opened 2020-10-21
Merging additional spectra variables


lgatto:
Adding new spectra variables is surprisingly difficult. Here's a detailed and reproducible example of why it is difficult and what we need:

## The test data

```r
> library("magrittr")
> library("Spectra")
> ## Load the raw data
> ms <- Spectra(msdata::proteomics(pattern = "2014", full.names = TRUE)) %>%
+     filterMsLevel(2L) %>%
+     dropNaSpectraVariables()
> ms
MSn data (Spectra) with 6103 spectra in a MsBackendMzR backend:
       msLevel     rtime scanIndex
     <integer> <numeric> <integer>
1            2   202.354       227
2            2   203.069       228
3            2   203.491       229
4            2   203.960       230
5            2   204.492       231
...        ...       ...       ...
6099         2   3600.47      7530
6100         2   3600.83      7531
6101         2   3601.18      7532
6102         2   3601.57      7533
6103         2   3601.98      7534
 ... 28 more variables/columns.

file(s):
TMT_Erwinia_1uLSike_Top10HCD_isol2_45stepped_60min_01-20141210.mzML.gz
Processing:
 Filter: select MS level(s) 2 [Wed Oct 21 09:56:29 2020] 
> spectraVariables(ms)
 [1] "msLevel"                 "rtime"                  
 [3] "acquisitionNum"          "scanIndex"              
 [5] "dataStorage"             "dataOrigin"             
 [7] "centroided"              "smoothed"               
 [9] "polarity"                "precScanNum"            
[11] "precursorMz"             "precursorIntensity"     
[13] "precursorCharge"         "collisionEnergy"        
[15] "isolationWindowLowerMz"  "isolationWindowTargetMz"
[17] "isolationWindowUpperMz"  "peaksCount"             
[19] "totIonCurrent"           "basePeakMZ"             
[21] "basePeakIntensity"       "ionisationEnergy"       
[23] "lowMZ"                   "highMZ"                 
[25] "injectionTime"           "filterString"           
[27] "spectrumId"              "scanWindowLowerLimit"   
[29] "scanWindowUpperLimit"   
```

## Easy case

It is straightforward to add a single new column of the right length.

```r
> ms$foo <- rnorm(length(ms))
> spectraData(ms)[, c("rtime", "foo")]
DataFrame with 6103 rows and 2 columns
         rtime        foo
     <numeric>  <numeric>
1      202.354  0.0686001
2      203.069 -0.8235955
[...]

PR #159, opened 2020-11-02
feat: add [[ and [[<-,MsBackend

jorainer:
- Add `[[` and `[[<-` methods for `MsBackend` (issue #149).
- Methods have a default implementation for `MsBackend` that use `$`, so the
  method does not necessarily have to be implemented for each backend.


R/MsBackend.R
```diff
@@ -1210,3 +1214,27 @@ setMethod("$", "MsBackend", function(x, name) {
...
+    do.call("$", list(x, i))
+})
+
+#' @exportMethod [[<-
+#'
+#' @rdname MsBackend
+setReplaceMethod("[[", "MsBackend", function(x, i, j, ..., value) {
+    if (!is.character(i))
+        stop("'i' is supposed to be a character defining the spectra ",
+             "variable to replace.")
```
lgatto, line 1236: to replace or create.

R/MsBackend.R
```diff
@@ -127,6 +127,10 @@
 #' - `$`, `$<-`: access or set/add a single spectrum variable (column) in the
 #'   backend.
 #'
+#' - `[[`, `[[<-`: access or set/add a single spectrum variable (column) in the
+#'   backend. The default implementation uses `$`, thus this methods don't have
```
lgatto, line 131: ... th**ese** methods don't have

lgatto (changes requested):

R/MsBackend.R
jorainer: Thanks, I've fixed that now

R/MsBackend.R
jorainer: Also here, thanks. I've fixed that now.

R/MsBackend.R
```diff
@@ -127,6 +127,10 @@
 #' - `$`, `$<-`: access or set/add a single spectrum variable (column) in the
 #'   backend.
 #'
+#' - `[[`, `[[<-`: access or set/add a single spectrum variable (column) in the
+#'   backend. The default implementation uses `$`, thus these methods don't have
```
sgibb, line 131: Shouldn't it be the other way around? Meaning: The default implementation of `$` uses `[[`.

`?"$"`
> Both ‘[[’ and ‘$’ select a single element of the list.  The main
     difference is that ‘$’ does not allow computed indices, whereas
     ‘[[’ does.  ‘x$name’ is equivalent to ‘x[["name", exact =
     FALSE]]’.  Also, the partial matching behavior of ‘[[’ can be
     controlled using the ‘exact’ argument.

R/MsBackend.R
```diff
@@ -127,6 +127,10 @@
...
+#'   to be implemented for new classes extending `MsBackend`.
```
sgibb, line 132: Did you mean `MsBackendDataFrame` here? cause `MsBackend::$` throws an error *is not implemented for class*.

R/MsBackend.R
jorainer: Yes, that would make more sense - only we required `$` and `$<-` for all backends that far, so all backends have these methods implemented, but not `[[` and `[[<-`. With the current implementation (`[[` is based on `$`) all backends gain this method without breaking anything.

R/MsBackend.R
jorainer: No, I meant it this way. I added a default method for `MsBackend`. Calling `[[` on a `MsBackend` will fail because `$` is not implemented/throws an error. But there is e.g. no need to implement `[[,MsBackendDataFrame` because it uses the `[[,MsBackend` which calls `$,MsBackendDataFrame`.

R/MsBackend.R
lgatto: @sgibb - is it OK to leave it like that, even if `[[` and `[[<-` should be those that `$` and `$<-` reply on. Implementation wise, it doesn't matter, as far as I see. It's historical in this case.

R/MsBackend.R
lgatto: @jorainer - I think the following are still missing `[[,Spectra` and `[[<-,Spectra`.

jorainer: Yes, makes sense to have also `[[,Spectra` and `[[<-,Spectra` implemented. I will add that.

jorainer: Just added the missing methods.

lgatto (approved):

lgatto:
I suggest to merge this. We can always get back to the open points if needed.

jorainer:
OK for me.

Issue #162, opened 2020-11-18
Bug in show,ProcessingStep


lgatto:
From `?ProcessingStep`:
```r
> ps <- ProcessingStep(sum)
> ps
Object of class "ProcessingStep"
 Function: Error in cat(" Function: ", object@FUN, "\n", sep = "") : 
  argument 2 (type 'builtin') cannot be handled by 'cat'
```

This however works:
```r
> executeProcessingStep(ProcessingStep("sum"), 1:10)
[1] 55
> ps <- ProcessingStep("sum")
> ps
Object of class "ProcessingStep"
 Function: sum
> executeProcessingStep(ps, 1:10)
[1] 55
```
The `show` method fails when `FUN` is a `function` given that is can't be coerced to a character. 

But that's not all - the same issue happens when on of the arguments in `@ARGS` is itself a function, for example when calling `filterIntensity(x, my_function)`. 

Here's a suggestion:

```r
setMethod("show", "ProcessingStep", function(object) {
    cat("Object of class \"", class(object), "\"\n", sep = "")
    if (is.function(object@FUN)) .FUN <- "FUN"
    if (is.character(object@FUN)) .FUN <- object@FUN
    cat(" Function: ", .FUN, "\n", sep = "")
    args <- object@ARGS
    if (length(args) > 0) {
        cat(" Arguments:\n")
        for (i in seq_along(args)) {
            if (is.function(args[[i]])) .ARG <- "FUN"
            else .ARG <- args[[i]]
            cat("  o ", names(args)[i], " = ",
                paste(.ARG, collapse = ", "), "\n", sep = "")
        }
    }
})
```

That would produce
```r
Object of class "ProcessingStep"
 Function: FUN
 Arguments:
  o intensity = FUN
  o msLevel = 1
``` 

Not sure if this is very helpful though.

jorainer:
Alternative suggestion:

```
> ps <- ProcessingStep(function(z) z + 4)
> show(ps)
Object of class "ProcessingStep"
 Function: function(z) z + 4
> ps <- ProcessingStep(sum, list(a = sqrt))
> show(ps)
Object of class "ProcessingStep"
 Function: function (..., na.rm = FALSE)  .Primitive("sum")
 Arguments:
  o  a  = function (x)  .Primitive("sqrt")
```

lgatto:
Sure, that's better.

Issue #70, opened 2019-09-26
show,ProcessingStep isn't working if FUN is a function


sgibb:
`show,ProcessingStep` throws an error if `@FUN` is a `function` (instead of `character`):

```r
ProcessingStep(sum)
# Object of class "ProcessingStep"
#  Function: Error in cat(" Function: ", object@FUN, "\n", sep = "") : 
#   argument 2 (type 'builtin') cannot be handled by 'cat'
```

in contrast:

```r
ProcessingStep("sum")
# Object of class "ProcessingStep"
#  Function: sum
```

jorainer:
Closing this issue, in _favour_ of #162

PR #163, opened 2020-11-19
fix: fix show,ProcessingStep (issue #162)

jorainer:
This fixes issue #162 (as well as issue #70)

### DIFF
--- a/R/ProcessingStep.R
+++ b/R/ProcessingStep.R
@@ -85,13 +85,14 @@ ProcessingStep <- function(FUN = character(), ARGS = list())  {
 #' @exportMethod show
 setMethod("show", "ProcessingStep", function(object) {
     cat("Object of class \"", class(object), "\"\n", sep = "")
-    cat(" Function: ", object@FUN, "\n", sep = "")
+    cat(" Function: ")
+    print(object@FUN)
     args <- object@ARGS
     if (length(args) > 0) {
         cat(" Arguments:\n")
         for (i in seq_along(args)) {
-            cat("  o ", names(args)[i], " = ",
-                paste(args[[i]], collapse = ", "), "\n", sep = "")
+            cat("  o ", names(args)[i], " = ")
+            print(args[[i]])
         }
     }
 })
--- a/R/ProcessingStep.R
+++ b/R/ProcessingStep.R
@@ -80,19 +80,23 @@ ProcessingStep <- function(FUN = character(), ARGS = list())  {
     new("ProcessingStep", FUN = FUN, ARGS = ARGS)
 }
 
+.cat_fun <- function(x) {
+    if (is.function(x))
+        "user-provided function"
+    else x
+}
+
 #' @rdname hidden_aliases
 #'
 #' @exportMethod show
 setMethod("show", "ProcessingStep", function(object) {
     cat("Object of class \"", class(object), "\"\n", sep = "")
-    cat(" Function: ")
-    print(object@FUN)
+    cat(" Function: ", .cat_fun(object@FUN), "\n", sep = "")
     args <- object@ARGS
     if (length(args) > 0) {
         cat(" Arguments:\n")
         for (i in seq_along(args)) {
-            cat("  o ", names(args)[i], " = ")
-            print(args[[i]])
+            cat("  o ", names(args)[i], " = ", .cat_fun(args[[i]]), sep = "")
         }
     }
 })
--- a/R/ProcessingStep.R
+++ b/R/ProcessingStep.R
@@ -96,7 +96,8 @@ setMethod("show", "ProcessingStep", function(object) {
     if (length(args) > 0) {
         cat(" Arguments:\n")
         for (i in seq_along(args)) {
-            cat("  o ", names(args)[i], " = ", .cat_fun(args[[i]]), sep = "")
+            cat("  o ", names(args)[i], " = ", .cat_fun(args[[i]]),
+                "\n", sep = "")
         }
     }
 })
--- a/tests/testthat/test_ProcessingStep.R
+++ b/tests/testthat/test_ProcessingStep.R
@@ -15,3 +15,13 @@ test_that("ProcessingStep executeProcessingStep", {
     ## Pass optional arguments to ...
     expect_identical(executeProcessingStep(ps, na.rm = TRUE), 10)
 })
+
+test_that("show,ProcessingStep works", {
+    ps <- ProcessingStep(sum, list(1:4))
+    show(ps)
+
+    ps <- ProcessingStep(function(z) z + 4)
+    show(ps)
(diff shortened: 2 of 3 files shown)


lgatto:
Here's a real-life example that isn't very readable. I don't think printing de function body is a good solution; `show` print a much shorter and readable, I think.

```r
> x <- Spectra(fl)
> ## function used as arguement to filterIntensity()
> low_int <- function(x, percentage = 0.05) {
+   x > max(x, na.rm = TRUE) * percentage
+ }
> x1 <- filterRt(x, rt = c(rtime(x)[1] - 24, rtime(x)[1] + 24))
> x1 <- filterEmptySpectra(x1)
> x1 <- filterMzRange(x1, c(1648, 1655))
> x1 <- combineSpectra(x1, intensityFun = sum)
Backend of the input object is read-only, will change that to an 'MsBackendDataFrame'
> x1 <- filterIntensity(x1, low_int)
> x1 <- pickPeaks(x1)
> x1@processingQueue
[[1]]
Object of class "ProcessingStep"
 Function: function(x, spectrumMsLevel, intensity,
                                             msLevel = spectrumMsLevel, ...) {



    if (!spectrumMsLevel %in% msLevel || !length(x))
        return(x)
    keep <- intensity(x[, "intensity"])
    if (!is.logical(keep) || length(keep) != nrow(x))
        stop("Error in filterIntensity: the provided function does not return ",
             "a logical vector of length equal to the number of peaks.",
             call. = FALSE)
    x[which(keep), , drop = FALSE]
}
<environment: namespace:Spectra>
 Arguments:
  o  intensity  = function(x, percentage = 0.05) {
  x > max(x, na.rm = TRUE) * percentage
}
  o  msLevel  = [1] 1

[[2]]
Object of class "ProcessingStep"
 Function: function(x, spectrumMsLevel, centroided = NA,
                        halfWindowSize = 2L, method = c("MAD", "SuperSmoother"),
                        snr = 0L, k = 0L, descending = FALSE, threshold = 0,
                        msLevel = spectrumMsLevel, ...) {
    if (!(spectrumMsLevel %in% msLevel))
        return(x)
    if (!nrow(x)) {
        warning("Spectrum is empty. Nothing to pick.")
        return(x)
    }

    n <- noise(x[, 1L], x[, 2L], method = method, ...)

    l <- localMaxima(x[, 2L], hws = halfWindowSize)

    p <- which(l & x[, 2L] > (snr * n))

    if (k > 0L) {
        cbind(mz = refineCentroids(x = x[, 1L], y = x[, 2L], p = p,
                                   k = k, threshold = threshold,
                                   descending = descending),
              intensity = x[p, 2L])
    } else {
        x[p, , drop = FALSE]
    }
}
<environment: namespace:Spectra>
 Arguments:
  o  halfWindowSize  = [1] 2
  o  method  = [1] "MAD"
  o  snr  = [1] 0
  o  k  = [1] 0
  o  descending  = [1] FALSE
  o  threshold  = [1] 0
  o  msLevel  = [1] 1
```

lgatto (commented):
See comment in the PR conversation.

jorainer:
Agree, that doesn't look good at all. I first tried to just get the function name with `deparse(substitute(` but that did not work for the arguments - there it was then simply printing `args[1]`.
So maybe your solution is better. I will change.

lgatto:
Maybe an exception, but for anonymous functions, there wouldn't be any name at all.

jorainer:
`show,ProcessingStep` will now print `"user-provided function"` if an argument or `@FUN` is a `function`.

tests/testthat/test_ProcessingStep.R
```diff
@@ -15,3 +15,20 @@ test_that("ProcessingStep executeProcessingStep", {
...
+test_that(".cat_fun works", {
+    res <- .cat_fun("sum")
+    expect_equal(res, "sum")
+    res <- .cat_fun(sum)
+    expect_equal(res, "user-provided function")
+})
+
+test_that("show,ProcessingStep works", {
+    ps <- ProcessingStep(sum, list(1:4))
+    show(ps)
```
sgibb, line 28: That's not a real test. This just fails if there is an error. What about `expect_output`?

sgibb (changes requested):
Nice fix but the unit tests for `show` don't really test anything.

tests/testthat/test_ProcessingStep.R
jorainer: :) you got me - I was just too lazy... I will fix.

lgatto:
I now get
```r
> x1@processingQueue
[[1]]
Object of class "ProcessingStep"
 Function: user-provided function
 Arguments:
  o intensity = user-provided function  o msLevel = 1
[[2]]
Object of class "ProcessingStep"
 Function: user-provided function
 Arguments:
  o halfWindowSize = 2  o method = MAD  o snr = 0  o k = 0  o descending = FALSE  o threshold = 0  o msLevel = 1
```
I think you are missing a newline in the `cat()`:
```r
        for (i in seq_along(args)) {
            cat("  o ", names(args)[i], " = ", .cat_fun(args[[i]]), "\n", sep = "")
        }
```

to get 
```r
> x1@processingQueue
[[1]]
Object of class "ProcessingStep"
 Function: user-provided function
 Arguments:
  o intensity = user-provided function
  o msLevel = 1

[[2]]
Object of class "ProcessingStep"
 Function: user-provided function
 Arguments:
  o halfWindowSize = 2
  o method = MAD
  o snr = 0
  o k = 0
  o descending = FALSE
  o threshold = 0
  o msLevel = 1
```

R/ProcessingStep.R
```diff
@@ -80,18 +80,23 @@ ProcessingStep <- function(FUN = character(), ARGS = list())  {
...
     cat("Object of class \"", class(object), "\"\n", sep = "")
-    cat(" Function: ", object@FUN, "\n", sep = "")
+    cat(" Function: ", .cat_fun(object@FUN), "\n", sep = "")
     args <- object@ARGS
     if (length(args) > 0) {
         cat(" Arguments:\n")
         for (i in seq_along(args)) {
-            cat("  o ", names(args)[i], " = ",
-                paste(args[[i]], collapse = ", "), "\n", sep = "")
+            cat("  o ", names(args)[i], " = ", .cat_fun(args[[i]]), sep = "")
```
lgatto, line 99: I think there's `\n` missing in your `cat()`

jorainer: yep, thanks. Will fix that - and add the correct unit tests.

jorainer:
I fixed now the missing `"\n"` and added *real* unit tests for `show,ProcessingStep`.

lgatto (approved):

jorainer:
I will also fix this in the release version

Issue #171, opened 2021-01-13
GNPS-similarity measure


jorainer:
GNPS considers also the precursor m/z of the spectra in the peak matching. Thus, to allow that, we would need to pass the precursor m/z of the spectra to the *join* function in `compareSpectra`.

jorainer:
Passing additional parameters `xPrecursorMz` and `yPrecursorMz` by the `.compare_spectra_chunk_prec` function to the `.peak_compare` function has no impact on performance (even not for `MsBackendMassbankSql` that retrieves data from the database).

jorainer:
GNPS uses a different way to match peaks between spectra. In addition to the *direct* matching peaks the algorithm considers peaks matching if their m/z difference is the same as the difference in the spectras' precursor m/z (please correct if I got it wrong @michaelwitting).

The first implementation is from @michaelwitting based on the R code from [this publication](https://pubs.acs.org/doi/10.1021/acs.analchem.0c02521)

```r
join_gnps <- function(x, y, xPrecursorMz, yPrecursorMz, tolerance = 0, ppm = 0) {
  
  # calculate differences of precursor
  precursorDiff <- yPrecursorMz - xPrecursorMz
  
  # create alignment matrix
  alignment <- data.frame(matrix(ncol = 4))
  colnames(alignment) <- c("mz.x", "int.x", "int.y","mz.y")
  
  h <- 1
  
  for (m in 1:nrow(x)){
    
    mz.diff1 <- abs(x[m,1] - y[,1])
    mz.diff2 <- abs(x[m,1] - y[,1] + precursorDiff)
    
    if(min(mz.diff1) <= tolerance){
      
      alignment[h,1:2] <- x[m,1:2]
      alignment[h,3] <- y[mz.diff1 == min(mz.diff1),2][1]
      alignment[h,4] <- y[mz.diff1 == min(mz.diff1),1][1]
      
      h <- h + 1
      
    }
    
    if(precursorDiff != 0) {
      
      if(min(mz.diff2) <= tolerance){
        
        alignment[h,1:2] <- x[m,1:2]
        alignment[h,3] <- y[mz.diff2 == min(mz.diff2),2][1]
        alignment[h,4] <- y[mz.diff2 == min(mz.diff2),1][1]
        
        h <- h + 1
        
      }
    }
  }
  
  alignment <- alignment[complete.cases(alignment),]
  
[...]

PR #178, opened 2021-02-16
Pass also precursor m/z to the similarity functions and add GNPS score

jorainer:
This PR adds the following changes:
- pass also precursor m/z values of the two spectra that are compared in `compareSpectra` to the spectra comparison function `FUN` (issue #171).
- add `joinPeaksGnps` and `gnps` functions to perform a peak matching and similarity score calculation as done by GNPS (issue #171).


R/peaks-functions.R
```diff
@@ -224,6 +224,20 @@ NULL
 #'   between `x` and `y` are returned by `type = "inner"`, i.e. only
 #'   peaks present in both spectra are reported.
 #'
+#' - `joinPeaksGnps` matches/maps peaks between spectra with an approach
+#'   similar to the one used in GNPS: peaks are considered matching if a) the
```
lgatto, line 228: Could you clarify what you mean by *similar*? Same algorithm but different implementation? Any tests that assess if identical or what the deviations are?

R/peaks-functions.R
```diff
@@ -286,11 +310,150 @@ NULL
...
+#'     first peak (row) of `y` or `NA_real_` if the peak does not match.
+#'
+#' @param y `numeric` peak `matrix` with m/z (first column) and intensity
+#'     values (second column). Rows in `x` and `y` are considered to contain
+#'     matching peaks, i.e. the first row in `x` contains a peak matching the
+#'     first peak (row) of `y` or `NA_real_` if the peak does not match.
+#'
+#' @param ... ignored
+#'
+#' @author Johannes Rainer, Michael Witting, based on the code from the ref
```
lgatto, line 396: ... based on code from Xing et al. (2020).

R/peaks-functions.R
```diff
@@ -286,11 +310,150 @@ NULL
...
+#' pmz_x <- 91
+#' pmz_y <- 105
+#'
+#' ## joinPeaksGnps finds 4 matches
+#' map <- joinPeaksGnps(x, y, pmz_x, pmz_y)
+#' map
+#'
+#' ## Calculate similarity
+#' gnps(map$x, map$y)
+gnps <- function(x, y, ...) {
```
lgatto, line 426: Should this function not rather be added to `MsCoreUtils`? It doesn't use or return `Spectra` objects.

R/peaks-functions.R
jorainer: It's actually the same approach but different implementation (GNPS is based on python). There are differences, but after the 3rd decimal.

R/peaks-functions.R
jorainer: I will update.

R/peaks-functions.R
lgatto: May be simply mention this in the documentation.

R/peaks-functions.R
jorainer: I originally also thought the same. But this function needs to be called in combination with `joinPeaksGnps` to get GNPS scores. Putting it into `MsCoreUtils` I'm afraid people might get the impression that this alone calculates GNPS similarity. But I have no strong opinion on that. I could also add it to `MsCoreUtils` if you think that's better.

R/peaks-functions.R
jorainer: Done

R/peaks-functions.R
jorainer: Done

R/peaks-functions.R
lgatto: Does that mean that the function on its own is likely to be useless? If so, them I would suggest to keep it in `Spectra` but not export it. If there is a use for it, then we a move to `MsCoreUtils` could be considered.

jorainer: Maybe I should explain  better: in order to calculate GNPS scores one would have to call:

```r
compareSpectra(sps, sps2, MAPFUN = joinPeaksGnps, FUN = gnps)
```

so we need to export `gnps`.  I could move `gnps` to `MsCoreUtils`, but then I would need also to add a function `join_gnps` there (that does the peak matching as in GNPS) and mention in the documentation that `gnps` only returns GNPS scores, if the peaks were first matched with `join_gnps`.

jorainer: OK, in the end I agree. I will add the `join_gnps` and `gnps` functions to the `MsCoreUtils` package and then update the PR here. So you will get another review request today over in `MsCoreUtils` :)

Issue #182, opened 2021-03-05
Pass arbitrary spectraVariables to the function applied by `addProcessing`


jorainer:
Using the `addProcessing` function it is possible to apply any user-provided function to the peak matrix. Currently, this is used by methods like `smooth` etc. This allows to manipulate m/z and intensity values of spectra on **any** backend (even if we have only a read-only backend).

Thus far we automatically pass also the spectra variables `msLevel` and `centroided` to the function which will be applied to each peak matrix. This is hardcoded into the function since most functions so far (`smooth`, `pickPeaks` etc) needed these parameters. It is however at present **not** possible to pass arbitrary spectra variables to the function (which would e.g. be needed for issue #181 ).

I think however I have a solution to this: I suggest that the user has the possibility to define with `addProcessing` which spectra variables should be made available the the function that should be applied to each peak matrix. These are then extracted and passed to the function on demand. This could be implemented with the following changes:

- refactor `addProcessing(FUN, ..., spectraVariables = character())`
- add a new slot `@processingQueueVariables` which can be used to store these requested `spectraVariables`.
- when `.peaksapply` is called, the requested spectra variables are extracted and passed along with the peak matrices to the queue functions.

The biggest change will be the addition of a new slot - we would have to provide a `updateObject` function that adds this slot if an old `Spectra` is loaded and we need to check for the presence of this slot in the `.peaksapply` and `addProcessing` functions.

I will implement this and compare its performance - but I don't expect a big loss in performance. I think however that this will improve the flexibility of `Spectra` considerably (and it will be easier for the user to provide custom functions).

@lgatto @sgibb I would really like to get your thoughts (and concerns?) about this.

lgatto:
Sounds good to me.

jorainer:
I've added some examples to the vignette describing this in detail.

jorainer:
Implemented and merged into master.

Issue #181, opened 2021-03-05
Efficient way to calculate neutral losses


jorainer:
Goal: subtract the precursor m/z values from all m/z values of a `Spectra`. 
Problem: `addProcessing` does not work as the precursor m/z is not passed to the peaks matrix processing functions.

Possible solutions:
- `mapply`-like function that allows to pass e.g. `precursorMz` and returns a `Spectra` (`endomapply`?).
- `addProcessing` gets also access to each row in `spectraData`, not only the peaks.

jorainer:
An ad-hoc possibility to do that is by replacing the m/z values like in the example below:

```r
df <- DataFrame(msLevel = c(1L, 2L, 2L, 2L), precursorMz = c(NA, 13, 12.2, 15),
                rtime = c(1, 2, 3, 4))
df$mz <- list(c(1:6), c(20, 24, 54), c(17, 200, 220, 221),
              c(123, 134, 135, 139))
df$intensity <- list(c(123, 343, 232, 45, 31, 532), c(5, 123, 99),
                     c(4324, 33499, 12, 5), c(151, 546, 1315, 24))
sps <- Spectra(df)

ms2 <- msLevel(sps) == 2L
mz(sps@backend)[ms2] <- mz(sps)[ms2] - precursorMz(sps)[ms2]
```

At present we have to call `mz(sps@backend)` because the `mz<-` is not (yet) implemented for `Spectra`. As a result we get then:

```r
> mz(sps)
NumericList of length 4
[[1]] 1 2 3 4 5 6
[[2]] 7 11 41
[[3]] 4.8 187.8 207.8 208.8
[[4]] 108 119 120 124
```

the m/z of the MS 1 spectrum were not changed while we subtracted the precursor m/z from each MS2 spectrum.

**Note**: this **only** works for backends that support replacing m/z and intensity values (such as the `MsBackendDataFrame`).

I will still think of a version that would allow such operations to be applied independently of the backend type.

michaelwitting:
Thank you very much. That helps. Any issues with the order of the mz values or is `mz(sps@backend)` taking care of this?

jorainer:
`mz(sps@backend)` **has** to take care of that. this depends on how the backend is implemented. For our backends it is guaranteed. So the code above has to work correctly.

michaelwitting:
Great, I would need this for some MS2 spectra, which are handled as `MsBackendDataFrame`. So everything should be fine. I will test with some data I have.

[...]

PR #183, opened 2021-03-08
Support arbitrary spectra variables in addProcessing

jorainer:
This PR adds the functionality to support passing arbitrary spectra variables to the user-provided function in `addProcessing` (issue #182). These can be specified with the parameter `spectraVariables` and these can then be accessed in the user-defined function. This enables e.g. to calculate neutral loss spectra by subtracting the precursor m/z from the m/z values of each spectrum (issue #181).


jorainer:
just pinging (or better bugging) you again with the PR review request :) @lgatto @sgibb

lgatto:
looking at it now

R/Spectra-functions.R
```diff
@@ -38,26 +59,63 @@ addProcessing <- function(object, FUN, ...) {
...
 #'
 #' @noRd
-.apply_processing_queue <- function(x, msLevel, centroided, queue = NULL) {
+.apply_processing_queue <- function(x, spectraData = data.frame(),
+                                     queue = NULL) {
     if (length(queue)) {
+        args <- list()
+        ls <- length(spectraData)
+        if (ls)
+            colnames(spectraData) <- sub("^msLevel$", "spectrumMsLevel",
```
lgatto, line 78: Why do we need that renaming? Can't we not just use `msLevel` instead of `spectrumMsLevel`?

R/Spectra.R
```diff
@@ -1684,7 +1692,8 @@ setMethod("filterIntensity", "Spectra",
                            "a lower and upper limit.")
                   object <- addProcessing(object, .peaks_filter_intensity,
                                           intensity = intensity,
-                                          msLevel = msLevel.)
+                                          msLevel = msLevel.,
```
lgatto, line 1695: Do we need to keep this? Can't we define that the default `spectraVariables = c("msLevel", "centroided")` and drop those single variable arguments? I see that `msLevel` shows up sometimes, but not `centroided`, so I might be missing something.

lgatto: Or, does `msLevel` refer to the MS levels on which the processing need to be applied? Then it makes sense.

R/Spectra-functions.R
lgatto: I now guess the reason is that `msLevel` refers to the level on which the processing is applied on and `spectrumMsLevel` the actual level of the spectra (from `spectraVariables`).

vignettes/Spectra.Rmd
```diff
@@ -478,11 +478,68 @@ lengths(sps_2)
...
+mz(sps_3)
+```
+
+As we can see, the precursor m/z was subtracted from each m/z of the respective
+spectrum. A better version of the function, that only calculates neutral loss
+spectra for MS level 2 spectra would be the `neutral_loss` function below. Since
+we are accessing also the spectrum's MS level we have to call `addProcessing`
+adding also the spectra variable `msLevel` to the `spectraVariables`
+parameter. Note however that the `msLevel` spectra variable **is by default
+renamed** to `spectrumMsLevel` prior passing it to the function. We have thus to
```
lgatto, line 517: I think this warrants some more explanations, but probably in the man page.

R/Spectra-functions.R
```diff
@@ -71,6 +129,9 @@ addProcessing <- function(object, FUN, ...) {
 #'     matrix will be passed to the first argument of the function. The function
 #'     should also have arguments `...`.
 #'
+#' @param spectraVariables `character` defining spectra variables that should
```
lgatto, line 132: Aren't there default parameters that are always passed? Or is that not the case anymore?

R/Spectra-functions.R
jorainer: Exactly. We basically wanted to be consistent with `MSnbase` and allow the user to specify the MS level on which to apply the function with parameter `msLevel`. That's why we renamed the spectrum's MS level variable to `spectrumMsLevel`.

R/Spectra.R
jorainer: Right, `msLevel` refers to the MS level on which the processing need to be applied. I know, this thing with `msLevel` and `spectrumMsLevel` is confusing...

vignettes/Spectra.Rmd
jorainer: Agree. I thought I have also added such information to the man page, but apparently I did not. Will do.

R/Spectra-functions.R
jorainer: now, by default, we're only passing the peak matrix. I would keep it to the user/developer that he/she defines which variables he/she needs in the function. that is also slightly faster and also simplifies the function to be applied to the peaks, because there is no longer the need that certain parameters (`spectrumMsLevel` and `centroided`) have to be present.

vignettes/Spectra.Rmd
jorainer: Done in https://github.com/rformassspectrometry/Spectra/commit/fadf42f1eb8bd3b2df4c32805af7333f5065a539

lgatto (approved):

PR #190, opened 2021-03-25
.isotope_peak function for isotope peak groups identification

andreavicini:


R/function-isotope.R
```diff
@@ -0,0 +1,149 @@
...
+  idxs_o <- which(x[, 2] > 0)
+  xred <- x[idxs_o, ]
+  if (is.null(seed_mz)) idxs <- 1:nrow(xred)
+  else idxs <- na.omit(MsCoreUtils::closest(seed_mz, xred[, 1], tolerance = 0,
+                                            ppm = ppm, duplicates = "closest"))
+  sgd <- NULL
+  for (i in idxs)
+  {
+    append_to_current <- FALSE
+    for (j in 1:nrow(isotopeDefinition))
```
jorainer, line 61: Do we need this loop? wouldn't it work if you did:

```r
tmp <- x[i, 1] + isotopeDefinition[, 1]
cls <- closest(tmp, x[, 1], tolerance = 0, ppm = ppm)
```

That way (I guess) you could also get around the `append_to_current`.

R/function-isotope.R
```diff
@@ -0,0 +1,149 @@
...
+        else
+        {
+          lst[[len]] <- c(lst[[len]], idxs_o[cls + i])
+          sgd[len] <- sgd[len] + tmp - xred[cls + i, 1]
+        }
+      }
+    }
+  }
+
+  if(length(lst) > 1 && duplicates == "closest")
```
jorainer, line 83: I was thinking if we need this at all or if we should simply avoid duplicates from the start. I think if we start with the smallest m/z and find peaks that match isotope conditions for one peak, these peaks (that were assigned to an isotope cluster) should not be used as "M" peak to search for possible matching isotope peaks. In other words, assume peak with index 4 is our M and peak 6 and 7 are matching the isotope rules, no isotope peaks should be searched for them and they should also not be considered as potential isotope peaks anymore.

What I would suggest is the following alternative approach:

```r
to_test <- rep(TRUE, nrow(x))
full_index <- seq_len(nrow(x))
while(!is.na(i <- match(TRUE, to_test))) {
    tmp <- x[i, 1] + isotopeDefinitions[, 1]
    idxs <- closest(tmp, x[to_test, 1]) # test only against peaks that are not yet in an isotope cluster
    ## BE AWARE THAT idxs IS NOW RELATIVE TO to_test. convert to "real" indices
    if (length(idxs)) {
        idxs <- full_index[to_test][idxs]
        ## check the intensity condition whether they fit
        ints <- x[idxs, 2]
        ...
        ## Disable testing of the M peak as well as of possible matching isotope peaks
        to_test[c(i, idxs[matched])] <- FALSE
    }
}
```

Maybe you find a better way than my approach above, as I guess that this continuous subsetting with `to_test` might have a negative effect on the performance.

Could be that that way we *might* loose some potential isotope peaks, but I would accept that because I think that should be a rare event. It will be however important to clearly explain this in the documentation.

jorainer (changes requested):
Thanks @andreavicini for the PR. Looks good, but I would suggest some changes (that might also eventually speed up the calculation).

Please also have a look at the code from @sgibb mentioned in https://github.com/rformassspectrometry/MetaboCoreUtils/issues/10

R/function-isotope.R
```diff
@@ -0,0 +1,149 @@
...
+#' @examples
+#'
+.isotope_peaks <- function(x, isotopeDefinition = isotopeDefinition(), ppm = 20,
+                           seed_mz = NULL, duplicates = "keep") {
+
+  if(duplicates != "keep" && duplicates != "closest" && duplicates != "remove")
+    stop("'duplicates' has to be one of \"keep\", \"closest\" or \"remove\".")
+
+  lst <- list()
+  idxs_o <- which(x[, 2] > 0)
```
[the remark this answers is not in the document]
andreavicini: Yes, if I remember correctly I asked you about a lower limit but since you said just put the condition intensity>0 I thought it was best not to run over those indexes in the first place (obtaining the same result). Moreover, in any case, in the example you provided almost half of the rows in the peak matrix had intensity = 0 so I thought in a situation like this it was convenient since we loop over fewer indexes and we don’t consider them when we call the closest function.
About the indexes, I will check it again but I thought I had made sure that the user receives the indexes of the original matrix and not those of the subsetted matrix.

R/function-isotope.R
andreavicini: Yes, I think that would work. I don't know why I did in that strange way. Thank you!

R/function-isotope.R
andreavicini: Yes, I can do this. Just to see it in an example with the example peak matrix you provided me and the simple two rows isotope definition we lose around 30 groups (over 133 that we obtain with keep parameter) if I'm not mistaken.

R/function-isotope.R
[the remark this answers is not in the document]
jorainer: I see your point. but generally intensities should be > 0, so I think you're not gaining a lot when skipping them. I would keep them in the initial implementation to keep the code simple.

R/function-isotope.R
jorainer: Uh, that's a lot. But this means that the 30 groups are all overlapping other groups, right? Maybe keep also your original function. I would like to see if we gain anything with this, i.e. how the performance of the two functions are on different examples.

andreavicini: There is a significant number of overlappings in the list of groups obtained with "keep" parameter. For example, when I use the parameter "closest" on the same example around 80 groups are removed. Yes, I’ll implement it so that maybe we can compare the results :)

jorainer (changes requested):
Thanks @andreavicini for the PR! Looks really good. I would suggest some minor changes/improvements.

R/function-isotope.R
```diff
@@ -0,0 +1,110 @@
...
+      tmp <- x[i, 1] + mzd
+      wtt <- which(to_test)
+      cls <- MsCoreUtils::closest(tmp, x[wtt, 1], tolerance = tolerance, 
+                                  ppm = ppm, duplicates = "closest")
+      int_ok <- .is_isotope_intensity_range(x[, 2][wtt[cls]], x[i, 1] * charge,
+                                            x[i, 2], substDefinition)
+      if (length(int_ok) > 0)
+      {
+        lst[[length(lst) + 1]] <- c(i, wtt[cls][int_ok])
+        to_test[wtt[cls][int_ok]] <- FALSE
```
sgibb, line 69:
```suggestion
        lst[[i]] <- c(i, wtt[cls][int_ok])
        idxs <- idxs[-cls[int_ok]]
```

R/function-isotope.R
```diff
@@ -0,0 +1,110 @@
...
+      }
+    }
+  }
+  lst
```
sgibb, line 73:
```suggestion
  lst[lengths(lst) > 0]
```

sgibb (changes requested):
Sorry I add a lot of comments. The general idea is to work on indicies or logical values only. Currently you have both and need to do the housekeeping which is valid/FALSE in both vectors. My suggestions just keep the indicies. Feel free to change that completely to logical but I would strongly suggest to use just one vector.

R/function-isotope.R
```diff
@@ -0,0 +1,110 @@
...
+#' @examples
+#'
+.isotope_peaks <- function(x, substDefinition = substDefinition(), 
+                           tolerance = 0, ppm = 20, seedMz = numeric(),
+                           charge = 1) {
+  lst <- list()
+  to_test <- rep(TRUE, nrow(x))
+  to_test[which(x[, 2] <= 0)] <- FALSE
+  if (length(seedMz)) 
+    idxs <- na.omit(MsCoreUtils::closest(seedMz, x[which(to_test), 1],
```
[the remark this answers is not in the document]
andreavicini: In your suggested change, to_test has to be replaced with `idxs`, right? And in general `to_test` is no longer required right?

andreavicini (commented):
Thanks @sgibb for your great suggestions!

R/function-isotope.R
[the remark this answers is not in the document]
sgibb: Indeed.

andreavicini:
@sgibb I realized that with the proposed use of `idxs`and removal of `to_test` we obtain a different behaviour than before in the case `seedMz` is provided by the user. Now the function would look only within the `idxs` (i.e. the indices of the peaks whose m/z value is closest to the `seedMz`) when looking for groups whose first peak matches `seedMz`, but the search should be extended also to all the other peaks. I think I should either add back a variable to specify which peaks to check or add an if-else to treat separately the two cases (when seedMz is provided and when it is not).

sgibb:
> @sgibb I realized that with the proposed use of `idxs`and removal of `to_test` we obtain a different behaviour than before in the case `seedMz` is provided by the user. Now the function would look only within the `idxs` (i.e. the indices of the peaks whose m/z value is closest to the `seedMz`) when looking for groups whose first peak matches `seedMz`, but the search should be extended also to all the other peaks. I think I should either add back a variable to specify which peaks to check or add an if-else to treat separately the two cases (when seedMz is provided and when it is not).

Sorry, I didn't thought about this. What about (code just for demonstration, doesn't work):
```r
idxs <- which(to_test)

if (length(seedMz))
  seedMz <- na.omit(closest(seedMz, x[idxs, 1], ...)
else
  seedMz <- idxs

# ...

for (i in seedMz) {
  # same code as before in the loop
}
```

andreavicini:
Thanks for the suggestion. I'm not sure I understood it correctly but I think if I did that and left the loop as before maybe wouldn't work when `seedMz` is given. In that case  for example `idx[1]` might be a low index while the indexes in seedMz might be greater and then the condition  `i == idxs[1]` would never be satisfied

sgibb:
> Thanks for the suggestion. I'm not sure I understood it correctly but I think if I did that and left the loop as before maybe wouldn't work when `seedMz` is given. In that case for example `idx[1]` might be a low index while the indexes in seedMz might be greater and then the condition `i == idxs[1]` would never be satisfied

Indeed you are right. Sorry for the confusion.

jorainer:
@andreavicini @sgibb , what is the current status of the review? Are your points all addressed now @sgibb ?

I also have a more fundamental question which is if we should add this functionality really to `Spectra` or maybe better to [MetaboCoreUtils](https://github.com/rformassspectrometry/MetaboCoreUtils) (or `MsCoreUtils` - but this approach is currently *tuned* towards small compound, i.e. metabolomics data). The reason is that we will use this function also in `xcms`, `MetaboAnnotation` and also `MsFeatures` - without actually using `Spectra` objects.

sgibb:
@jorainer I am not sure that we have a working version now (cause of my confusing last comments).
You are right this should go to `MsCoreUtils` or `MetaboCoreUtils`. If it is really _tuned_ to metabolomics I would vote for `MetaboCoreUtils`.

jorainer:
OK, I would then suggest the following: @andreavicini , please make a pull request adding this code to `MetaboCoreUtils` and @sgibb , if possible, I would then ask you to re-review the PR again there.

jorainer:
Closing this PR because (as agreed above) this fits better into `MetaboCoreUtils` (-> PR https://github.com/rformassspectrometry/MetaboCoreUtils/pull/38)

Issue #192, opened 2021-04-03
Adding a filterPrecursorCharge() filter function?


lgatto:
@jorainer - any objection?

jorainer:
Sounds good to me - will it filter on a range or a single value?

lgatto:
Range would be better indeed.

jorainer:
Then it's similar to the `filterPrecursorMz`. Maybe you could also add the generic to `ProtGenerics` then?

lgatto:
Sure, will do.

lgatto:
I have this implementation working:

```r
> library(rpx)
> px <- PXDataset("PXD000001")
> f <- pxget(px, "TMT_Erwinia_1uLSike_Top10HCD_isol2_45stepped_60min_01-20141210.mzML")
Loading TMT_Erwinia_1uLSike_Top10HCD_isol2_45stepped_60min_01-20141210.mzML from cache.
> library(Spectra)
> sp <- Spectra(f)
> table(msLevel(sp), precursorCharge(sp))
   
       2    3    4    5    6
  1    0    0    0    0    0
  2 3749 1955  368   25    6
> sp2 <- filterPrecursorCharge(sp, 2:3)
> table(msLevel(sp2), precursorCharge(sp2))
   
       2    3
  2 3749 1955
```
which however leads me to question the result, which also applies to `filterPrecursorMz` (and possibly others). 

@jorainer  We loose all MS1 spectra, given that we keep those that have a `precursorCharge`. Shouldn't we also keep all other MS levels?

jorainer:
I think this depends on the main purpose/use case for the function. I use `filterPrecursorMz` mostly for spectra matching where I only want to compare the query spectrum against target (reference) spectra with matching precursor m/z. Thus I actually want to drop all MS1 spectra or those MS2 spectra without precursor m/z. To be consistent it would make sense to have the same behaviour for `filterPrecursorCharge`. 

I think this is also the behaviour users might expect from such a function: if I want to filter by precursor, I will only get those for which the condition (precursor m/z or charge) matches - I might be confused if I get also MS1 spectra in return as they don't have a precursor. But that's only my opinion, happy to discuss.

lgatto:
I agree with you, but this is trivial:

```r
x %>%
   filterMsLevel(2L) %>%
   filterPrecursorMz(c(500, 501))
```
instead of 
```r
x %>%
   filterPrecursorMz(c(500, 501))
```

[...]

PR #193, opened 2021-04-06
feat: new filterPrecursorCharge() method - see issue #192

lgatto:


DESCRIPTION
```diff
@@ -8,25 +8,25 @@ Description: The Spectra package defines an efficient infrastructure
    backends tuned for fast data access and processing and backends for very
    large data sets ensuring a small memory footprint.
 Authors@R: c(person(given = "RforMassSpectrometry Package Maintainer",
-		    email = "maintainer@rformassspectrometry.org",
```
jorainer, line 11: you need also to bump the version above to 1.1.19

NEWS.md
```diff
@@ -2,6 +2,10 @@
 
 ## Changes in 1.1.18
```
jorainer, line 3:
```suggestion
## Changes in 1.1.19
```

jorainer (changes requested):
Looks good @lgatto ! Thanks for the PR - only change request is to bump the version of the package. After that you're free to merge.

DESCRIPTION
lgatto: Ah yes, forgot that! Done now, thanks

NEWS.md
lgatto: fixed too

Issue #202, opened 2021-05-06
Estimate precursor intensities for MS2 spectra


jorainer:
This is coming from https://github.com/jorainer/SpectraTutorials/issues/6 . Some instruments don't report precursor intensities, but it's straight forward to calculate them based on MS1 spectra data. The functionality for `OnDiskMSnExp` objects is currently implemented in `xcms::estimatePrecursorIntensity` - we should also implement that for `Spectra`.

jorainer:
Done

PR #203, opened 2021-05-07
feat: add estimatePrecursorIntensity (issue #202)

jorainer:
This PR adds a function `estimatePrecursorIntensity` that calculates precursor intensities for MS2 spectra based on matching MS1 peaks from the closest measured MS1 spectra.


lgatto:
Looking forward to reviewing that one :-)

R/Spectra-functions.R
```diff
@@ -623,4 +588,91 @@ joinSpectraData <- function(x, y,
...
+        x <- x[idx]
+    } else idx <- integer()
+    pmz <- precursorMz(x)
+    pmi <- rep(NA_real_, length(pmz))
+    idx_ms2 <- which(!is.na(pmz) & msLevel(x) == msLevel)
+    x_ms1 <- filterMsLevel(x, msLevel = msLevel - 1L)
+    pks <- .peaksapply(x_ms1)
+    for (i in idx_ms2) {
+        rt_i <- rtime(x[i])
+        before_idx <- which(rtime(x_ms1) < rt_i)
```
sgibb, line 637: Do we use something like this often? If that is the case I am going to implement a `which.first`/`which.last` function in `MsCoreUtils`? What do you think?

R/Spectra.R
```diff
@@ -1015,6 +1028,21 @@ NULL
 #' res <- lapply(intensity(sciex_im[1:20]), mean)
 #' head(res)
 #'
+#' ## Calculating the precursor intensity for MS2 spectra:
+#' ##
+#' ## Some MS instrument manufacturer don't report the precursor intensities
+#' ## for MS@ spectra. The `estimatePrecursorIntensity` function can be used
```
sgibb, line 1034:
```suggestion
#' ## for MS2 spectra. The `estimatePrecursorIntensity` function can be used
```

sgibb (approved):
I recognize that I didn't know the rest of the code very well anymore. But IMHO nearly everything is fine here. Just minor things.

R/Spectra-functions.R
```diff
@@ -623,4 +588,91 @@ joinSpectraData <- function(x, y,
...
+        warning("Estimation of precursor intensities for MS levels > 2 is ",
+                "not yet validated")
+    if (!length(x))
+        return(numeric())
+    if (is.unsorted(rtime(x))) {
+        idx <- order(rtime(x))
+        x <- x[idx]
+    } else idx <- integer()
+    pmz <- precursorMz(x)
+    pmi <- rep(NA_real_, length(pmz))
```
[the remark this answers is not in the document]
jorainer: It's a little counter intuitive, but:

```r
> library(microbenchmark)
> microbenchmark(rep(NA_real_, 1000), rep_len(NA_real_, 1000))
Unit: microseconds
                    expr min  lq  mean median   uq  max neval cld
     rep(NA_real_, 1000) 3.7 4.0 4.588    4.2 4.40 23.1   100  a 
 rep_len(NA_real_, 1000) 4.9 5.2 6.234    5.4 5.75 45.6   100   b
```

That's the reason I usually use `rep` instead of `rep_len`.

R/Spectra-functions.R
jorainer: Hm, actually, getting the index of the last (x) retention time(s) before and the first (x) after might indeed be something we could use more frequently.

jorainer:
Thanks for the review @sgibb ! Really always appreciate your reviews! I'll fix the CI and merge the PR (without your suggestion to use `rep_len` instead of `rep` (since performance is slightly better for the latter).

R/Spectra-functions.R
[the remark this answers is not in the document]
sgibb: Thanks, for the benchmark. Even `rep.int` is slower than `rep`, really counter intuitive.

R/Spectra-functions.R
sgibb: That's a nice small function. I am going to take that one: https://github.com/rformassspectrometry/MsCoreUtils/issues/85 in the next days.

jorainer: cool!

Issue #194, opened 2021-04-07
Bug in filterPrecursorScan()


lgatto:
```r
> library(rpx)
> PXD022816 <- PXDataset("PXD022816")
> (mzmls <- pxget(PXD022816, grep("mzML", pxfiles(PXD022816))[1:2]))
Loading QEP2LC6_HeLa_50ng_251120_01-calib.mzML from cache.
Loading QEP2LC6_HeLa_50ng_251120_02-calib.mzML from cache.
[1] "~/.cache/rpx/207e04bccd3dd_QEP2LC6_HeLa_50ng_251120_01-calib.mzML"
[2] "~/.cache/rpx/207e03973c908_QEP2LC6_HeLa_50ng_251120_02-calib.mzML"
> sp <- Spectra(mzmls)
> sp2 <- filterPrecursorScan(sp, 2490)
> length(sp2)
[1] 12
```

```r
spectraData(sp2)[, c("msLevel", "acquisitionNum", "precScanNum", "dataOrigin")]
```
| msLevel| acquisitionNum| precScanNum|dataOrigin                                           |
|-------:|--------------:|-----------:|:----------------------------------------------------|
|       1|           2479|          NA|207e04bccd3dd_QEP2LC6_HeLa_50ng_251120_01-calib.mzML |
|       2|           2480|        2479|207e04bccd3dd_QEP2LC6_HeLa_50ng_251120_01-calib.mzML |
|       1|           2482|          NA|207e04bccd3dd_QEP2LC6_HeLa_50ng_251120_01-calib.mzML |
|       2|           2485|        2482|207e04bccd3dd_QEP2LC6_HeLa_50ng_251120_01-calib.mzML |
|       1|           2486|          NA|207e04bccd3dd_QEP2LC6_HeLa_50ng_251120_01-calib.mzML |
|       2|           2490|        2486|207e04bccd3dd_QEP2LC6_HeLa_50ng_251120_01-calib.mzML |
|       1|           2479|          NA|207e03973c908_QEP2LC6_HeLa_50ng_251120_02-calib.mzML |
|       1|           2480|          NA|207e03973c908_QEP2LC6_HeLa_50ng_251120_02-calib.mzML |
|       2|           2482|        2480|207e03973c908_QEP2LC6_HeLa_50ng_251120_02-calib.mzML |
|       1|           2485|          NA|207e03973c908_QEP2LC6_HeLa_50ng_251120_02-calib.mzML |
[...]

PR #206, opened 2021-07-01
fix: per-file filtering in filterPrecursorScan (issue #194)

jorainer:
- Add parameter `f = dataOrigin(object)` to `filterPrecursorScan` to allow
  filtering within each sample/file. This fixes issue #194.


sgibb (approved):
I think the code is fine. Would wait for @lgatto's tests.

lgatto:
Yes, I confirm that this fixes the bug descibed in #194. Thanks !

lgatto:
Also pushed to Bioc.

PR #214, opened 2021-09-02
feat: new plotMzDelta functions

lgatto:


NAMESPACE
```diff
@@ -104,6 +106,7 @@ exportMethods(spectraVariables)
 exportMethods(spectrapply)
 exportMethods(split)
 exportMethods(tic)
+import(BiocParallel)
```
jorainer, line 109: could you please change that to a more selective import? We're already importing `BiocParallel::bpparam` and `BiocParallel::SerialParam`, so maybe this is even not needed.

R/mz-delta-functions.R
```diff
@@ -0,0 +1,213 @@
...
+#' proteomics data. Proteomics. 2011;11:
+#' 2182-2194. http://dx.doi.org/10.1002/pmic.201000602
+#'
+#' @param object An instance of class `Spectra()`.
+#'
+#' @param percentage `numeric(1)` between 0 and 1 indicating the
+#'     percentage of the most intense peaks in each MS2 spectrum to
+#'     include in the calculation. Default is 0.2.
+#'
+#' @param xlim `numeric(2)` with the upper and lower M/Z to be used to
```
jorainer, line 48: Maybe better to call that parameter `mzRange` instead? I find `xlim` a little unintuitive here, because this function is not plotting anything.

jorainer (changes requested):
Some minor change requests, otherwise I'm fine with it. Thanks @lgatto .

NAMESPACE
lgatto: Fixed in latest commit.

R/mz-delta-functions.R
lgatto: Changed in latest commit.

jorainer (approved):
Thanks @lgatto . All fine now.

lgatto:
Thank you for the review!

lgatto:
Also pushed to Bioc.

PR #215, opened 2021-09-06
features: joinSpectraData() checks for key duplicates

lgatto:
The `joinSpectraData(x, y, by.x, by.y)` function now checks for duplicates in the x and y keys. 

- If there's duplicates in the `Spectra` key (i.e. in `x[[by.x]]`), then an error is thrown. Indeed, this key should be unique and the user would need to sort this out it it wasn't. 
- If there's duplicates in the `y` key (i.e. in `y[[by.y]]`), then a warning is thrown, and the last occurrences of the duplicated entries are used (as as result of using `match()`). The documentation provides guidance how to manage this.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,8 @@
 
 ## Changes in 1.3.9
 
-- Nothing yet.
+- New features: `joinSpectraData()` now check for duplicated keys in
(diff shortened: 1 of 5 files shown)


jorainer (approved):
Thanks @lgatto ! Looks all fine to me.

PR #222, opened 2021-11-20
New countIdentifications() function

lgatto:


lgatto:
Hi @sgibb and @jorainer 

This draft PR add a new function, `nSequences()` that takes a proteomics Spectra object with identification results (as a `sequence` spectra variable). It then counts, for each scan (MS1 and MS2) the number of identification that results from that scan. This is obvious and uninteresting for MS2 scans, as these are either 0 or one (depending whether `sequence` is NA or not). But for MS1 scan, the function could how many identifications where found for in the descendant MS2 scans. This can in turn be used to generate a plot as shown below

image

In addition to your review, I have some questions/comments:
1. Ok to add this function in it's own file `nSequences.R`.
2. I'm not convinced by the name of the function.
3. The function is quite slow. How could it be improved? I suspect that the current implementation of `.filterSpectraHierarchy()` is slow for such a use-case, where it is called for all MS1 scans. I can think of a vectorised and/or a C-level implementation. What do you think?

R/nSequences.R
```diff
@@ -0,0 +1,142 @@
...
+#'
+#' ## Count number of sequences/identifications per scan
+#' sp <- nSequences(sp)
+#'
+#' ## MS2 scans either lead to an identification (5 instances) or none
+#' ## (76). Among the five MS1 scans in the experiment, 3 lead to MS2
+#' ## scans being matched to no peptides and two MS1 scans produced two
+#' ## and three PSMs respectively.
+#' table(sp$nSequences, sp$msLevel)
+nSequences <- function(object,
```
jorainer, line 105: Using a variable/name `sequence` for an annotation seems to me not straight forward - in small molecule MS we don't have sequences but chemical formulas or similar. Thus, what about *generalizing* this and using the term *annotation* instead of *sequence* and calling this function `countAnnotations`?

R/nSequences.R
```diff
@@ -0,0 +1,142 @@
...
+#' MS2 scans originating from any MS1 peak that lead to an
+#' identification.
+#'
+#' This function can be used to generate id-annotated total ion
+#' chromatograms, as can illustrated
+#' [here](https://rformassspectrometry.github.io/docs/sec-id.html#an-identification-annotated-chromatogram).
+#'
+#' @details
+#'
+#' The computer number of identifications is stored in a new spectra
```
jorainer, line 18: `computed number` not `computer number`

jorainer (commented):
That's an interesting function that might be useful also from a quality assessment perspective. I have just one comment/suggestion regarding to the function name.

Regarding your questions:
1) yes, absolutely fine to have the function in a separate .R file.
2) me neither - that's why I suggested to call it `countAnnotations` instead.
3) I would expect this function to be only called once on a `Spectra`, so I don't know how problematic performance is. I'm OK if you want to give a C-level implementation a go, but I'll not be a big of a help there...

R/nSequences.R
lgatto: thanks, fixed in #58a6e50b0e63e849e0d0fd6ca76aff238b125678

lgatto:
Yes, let's generalise the name - I'm happy to hear this would also be useful for metabolomics data. But I'm not convinced by *annotation*. In proteomics, you could have an annotation that doesn't equate to a peptide sequence (i.e. the identification) - for example the presence of an ammonium ion. 

What about using the term *identification* rather than sequence, which is just as good or a match in proteomics and seems to also fit metabolomics.

If you are happy with this, I would rename the function `countIdentifications()`.

jorainer:
Yes. `identification` is OK for me.

sgibb (changes requested):
Dear @lgatto that's a really nice addition and graphical representation!
Unfortunately I don't have an idea to really improve `.filterSpectraHierarchy`(it was meant to filter one single hierarchy) beside reconstructing the whole tree (which may take also some time) and count its leafs.
I just have question regarding the naming of `res`.

R/countIdentifications.R
```diff
@@ -0,0 +1,141 @@
...
+    ## Iterate over MS1 acquition numbers
+    N <- acquisitionNum(filterMsLevel(object, 1))
+    for (i in seq_along(N)) {
+        ## curent MS1 acquisition number
+        an <- N[i]
+        an_char <- as.character(an)
+        ## get logical with all descendant acquisitions
+        j <- .filterSpectraHierarchy(object$acquisitionNum,
+                                     object$precScanNum, an)
+        res[an_char] <- sum(!is.na(object[[identification]][j]))
```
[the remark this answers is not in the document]
lgatto: Here's an example that illustrates this, with a slice of data from acquisition numbers 101 to 118.

```r
> spdf
   msLevel acquisitionNum precScanNum sequence
1        1            101          NA     <NA>
2        2            102         101  PEPTIDE
3        2            103         101  PEPTIDE
4        2            104         101  PEPTIDE
5        2            105         101  PEPTIDE
6        2            106         101  PEPTIDE
7        1            107          NA     <NA>
8        2            108         107  PEPTIDE
9        2            109         107  PEPTIDE
10       2            110         107  PEPTIDE
11       2            111         107  PEPTIDE
12       2            112         107  PEPTIDE
13       1            113          NA     <NA>
14       2            114         113  PEPTIDE
15       2            115         113  PEPTIDE
16       2            116         113  PEPTIDE
17       2            117         113  PEPTIDE
18       2            118         113  PEPTIDE
> object <- Spectra(spdf)
> identification  <- "sequence"
> res <- as.integer(!is.na(object[[identification]]))
> names(res) <- acquisitionNum(object)
> N <- acquisitionNum(filterMsLevel(object, 1))
> i <- 1
> an <- N[i]
> an_char <- as.character(an)
> an
[1] 101
> an_char
[1] "101"
> j <- Spectra:::.filterSpectraHierarchy(object$acquisitionNum, object$precScanNum, an)
> j
 [1]  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE FALSE FALSE FALSE FALSE FALSE FALSE
[13] FALSE FALSE FALSE FALSE FALSE FALSE
> sum(!is.na(object[[identification]][j])) ## should be 5
[1] 5
> res[an]
<NA> 
  NA 
> res[an_char]
101 
  0 
```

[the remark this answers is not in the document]
sgibb: Point taken. I am fine with merging this PR.

sgibb (approved):

lgatto:
@jorainer - should we merge the two accepted PRs despite of the errors (as noted [here](https://github.com/rformassspectrometry/Spectra/pull/225#issuecomment-980654517))

jorainer:
Yes, I would do so. I will merge and ensure that in the other PR news/versions etc get updated too.

jorainer:
I will also add an intermediate fix until `mzR` is (finally - might still need some time) updated.

lgatto:
And what about pushing to Bioc? Go ahead or do we need to wait?

jorainer:
I will then push Spectra to BioC (once we have both PR in)

jorainer:
Done.

Issue #186, opened 2021-03-23
External backends running unit tests of Spectra?


jorainer:
This comes from an idea/suggestion from @sneumann at the *metaRbolomics* hackathon 2021:
We have `MsBackend` implementations in external packages. There is currently no way that we can ensure that these *work properly* or return the data in the expected format. I see there two solutions:

1) we leave this to the developers of the external backend, i.e. it's their responsibility to add corresponding unit tests in their package.
2) we have some unit tests (in `Spectra`) that can be called from external packages. No idea if and how this is working, but the idea would be that these unit tests allow to provide a `MsBackend` as input  and the test performs the most critical tests, such as:
    - check that the returned/provided data types are correct.
    - check that subsetting works
    - check that it is possible to add spectra variables
    - ?

Any thoughts, ideas welcome.

sgibb:
> 1. we leave this to the developers of the external backend, i.e. it's their responsibility to add corresponding unit tests in their package.

I guess that's the way we have to go because everything else would be very hard.

Regarding the second solution: we could provide an example mzML and a template testthat file where developers just have to plugin their backend and previously import the example mzML into their backend source. Or we could provide a function e.g.:
```r
testSpectraBackend(Spectra(
  source = "ourExampleConvertedFromMzMlIntoExternalBackend", 
  backend = MsBackend())) {
  # check mz/intensity values
  stopifnot(
    all(mz(spectra) == c(1, 2, 3, 4, 5)),
    all(mz(spectra) == c(1, 2, 3, 4, 5))
  )
  # ...
}
```
I guess no user ever calls `testthat::test_package()`. But if we provide such a function and promote it in the vignette, e.g.

> always test your backend before
```r
library("Spectra")
library("MsBackendExternal")
data("ExampleSpectrum", package = "MsBackendExternal")
testSpectraBackend(Spectra(ExampleSpectrum, backend = MsBackendExternal))
# TRUE
# or
# Error: mz is not 1:5 or whatever
```

sneumann:
I would still be optimistic that there is a way for 2.
we'd need `testthat::test_dir(Spectra/tests/backendtests)` and tests in there 
need to get the backend and file in that format as a variables or options. 
Yours, Steffen

hechth:
So we created some framework and think that defining an abstract set of tests based on a global variable (the backend) and the expected output. The backends can then call the tests and before that define the global variable to be the initialized backend.

sneumann:
[...]

PR #224, opened 2021-11-22
Install test suite for backend packages

sneumann:
This is working on #186 . Spectra now includes (and installs) a directory with tests that can be called from the backends. There could (should!) be a few different profiles, checking a (minimum) set of functionality provided by the backends. The initially provided one barely tests anything yet. Yours, Steffen


jorainer:
This PR looks extremely interesting, thanks @sneumann and @hechth. I would however provide an already initialized `MsBackend` with the `be` variable since not all backends will have (only) a `file` parameter (e.g. expecting a database connection instead or even a data frame or similar).

Another very important test suite would be a general `test_valid_MsBackend` test suite that takes an already initialized `MsBackend` variable `be` and then runs all methods defined in the `MsBackend.R` file and checks that the returned result is as expected (i.e. expected data type, length of data matches expectation etc). Question is if you are planning to add that too or if we (I) should do that?

inst/test_msmslibraries/test_MsBackendMsmslibraries.R
```diff
@@ -0,0 +1,20 @@
+## assuming an object "be" exists with an initialises backend
```
jorainer, line 1: Please add here a little more documentation/notes, such as
- what is expected (variable name(s), what they should be
- how can this test be called from another package
- what is being tested by the tests within this file

jorainer (changes requested):
I think that's an excellent contribution, thanks @sneumann and @hechth. I made some comments and in general I think it would be better to structure the directories in a different way:

inst/test_backends/ being the base folder, and in this one having different folders e.g. your test_msmslibraries/ defining different test suites.

sneumann:
I think this can now be closed in favour of https://github.com/rformassspectrometry/Spectra/pull/227 ?

jorainer:
It's up to you @sneumann - you're more than welcome to add additional test suites. #227 adds *only* test suites for accessing/setting variables.

jorainer:
Closing this PR now.

PR #231, opened 2021-12-10
Ms backend cached

jorainer:
This PR adds a `MsBackend` that provides a caching mechanism allowing to add or *modify* spectra variables of a read-only `MsBackend`. Spectra variable setter methods add values to a local slot with a `data.frame`, accessor methods first evaluate if a variable is available in the local cache and in case returns that.

Database-backed `MsBackend`s can extend this `MsBackendCached` instead of the `MsBackend` to have the caching mechanism available. Prime examples that can immediately use this backend are the  `MsBackendMassbankSql` and the `MsBackendCompDb`.


R/MsBackendCached.R
```diff
@@ -0,0 +1,616 @@
...
+    if (is.null(msg)) TRUE
+    else msg
+})
+
+#' @rdname MsBackendCached
+setMethod("backendInitialize", "MsBackendCached",
+          function(object, data = data.frame(), nspectra = 0L,
+                   spectraVariables = character(), ...) {
+              if (nrow(data) == 0L)
+                  data <- data.frame(matrix(ncol = 0, nrow = nspectra))
```
sgibb, line 285: or `data.frame(row.names = seq_len(nspectra))`

R/MsBackendCached.R
```diff
@@ -0,0 +1,616 @@
...
+        }
+        if (length(core_cols))
+            tmp <- DataFrame(lapply(.SPECTRA_DATA_COLUMNS[core_cols],
+                                    function(z, n) rep(as(NA, z), n), length(x)))
+        else tmp <- make_zero_col_DFrame(x@nspectra)
+        tmp$mz <- mzvals
+        tmp$intensity <- intvals
+        if (length(res))
+            res <- cbind(res, tmp)
+        else res <- tmp
```
sgibb, line 342: I don't really get this part. Why adding mz/intensity to `res` when it is always `NULL` or a `NumericList` with an `integer(0)` per row. Is this useful at all?

R/MsBackendCached.R
```diff
@@ -0,0 +1,616 @@
...
+
+#' @rdname MsBackendCached
+setMethod(
+    "spectraData", "MsBackendCached",
+    function(object, columns = spectraVariables(object)) {
+        .spectra_data(object, columns = columns)
+    })
+
+#' @rdname MsBackendCached
+setReplaceMethod("spectraData", "MsBackendCached",function(object, value) {
```
sgibb, line 363:
```suggestion
setReplaceMethod("spectraData", "MsBackendCached", function(object, value) {
```

R/MsBackendCached.R
```diff
@@ -0,0 +1,616 @@
...
+    if (missing(i))
+        return(x)
+    slot(x, "localData", check = FALSE) <- x@localData[i, , drop = FALSE]
+    x@nspectra <- nrow(x@localData)
+    x
+})
+
+#' @rdname MsBackendCached
+setMethod("$", "MsBackendCached", function(x, name) {
+    if (!any(spectraVariables(x) == name))
```
sgibb, line 378:
```suggestion
    if (all(spectraVariables(x) != name))
```

R/MsBackendCached.R
```diff
@@ -0,0 +1,616 @@
...
+        validObject(object)
+        object
+    })
+
+#' @rdname MsBackendCached
+setMethod("show", "MsBackendCached", function(object) {
+    n <- object@nspectra
+    cat(class(object), "with", n, "spectra\n")
+    if (n) {
+        idx <- union(1:min(6, n), max(1, n-5):n)
```
sgibb, line 432: `union` is just `unique(c(x, y))`
```suggestion
        idx <- unique(c(1L:min(6L, n), max(1L, n-5L):n))
```

sgibb (changes requested):
Thanks for this PR. I have some questions regarding the *mz*/intensity columns in `spectraData`.

R/MsBackendCached.R
```diff
@@ -0,0 +1,616 @@
...
+#'
+#' @author Johannes Rainer
+#'
+#' @name MsBackendCached
+#'
+#' @return See documentation of respective function.
+#'
+#' @seealso [MsBackend] for the documentation of MS backends.
+#'
+#' @md
```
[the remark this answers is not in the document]
jorainer: Nope. I will remove that.

R/MsBackendCached.R
```diff
@@ -0,0 +1,616 @@
...
+#'
+#' @exportClass MsBackendCached
+NULL
```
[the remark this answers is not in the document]
jorainer: Don't know - I always used that so far.

R/MsBackendCached.R
```diff
@@ -0,0 +1,616 @@
...
+
+setClass(
+    "MsBackendCached",
+    contains = "MsBackend",
+    slots = c(
+        spectraVariables = "character",
+        localData = "data.frame",
```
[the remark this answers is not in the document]
jorainer: The reason is that subsetting a `DataFrame` is not efficient (or at least way slower than `data.frame`). I don't see any big reason to have a `DataFrame` here - and for performance reasons I perfer now `data.frame` over `DataFrame` for most new code.

R/MsBackendCached.R
jorainer: That's nice. Thanks for the hint.

R/MsBackendCached.R
jorainer: So you mean why having the `tmp$mz <- mzvals` line when `mzvals` is `NULL`?

The reasoning of that code is:
- if there is any column `"mz"` (or `"intensity"`) requested, then initialize the variabl `mzvals` and **remove** the requested column.
- create a `DataFrame` with the requested columns and `NA` values of the correct data type.
- add `$mz` to this data frame. If it was not requested `NULL` (nothing) will be added. Otherwise the `NumericList` is added. This is to avoid having yet another `if` statements for mz and intensity.
 
This whole code just takes care to add *default* (empty) values for core spectra variables.

R/MsBackendCached.R
jorainer: Isn't `any` faster than `all` because it breaks when the first `TRUE` is returned?

R/MsBackendCached.R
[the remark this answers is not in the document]
jorainer: Actually, I don't want the `MsBackendCached` class (slots) etc to be documented/included in the roxygen documentation. We didn't do that for any class in `Spectra`.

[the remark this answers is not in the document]
sgibb: Fine!

R/MsBackendCached.R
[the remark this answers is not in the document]
sgibb: Ok.

R/MsBackendCached.R
sgibb: Yes, I guess you are right. I find `all( != )` easier to read and thought it would not be a time critical step. But you can keep `any`.

sgibb (approved):
Fine for me!

R/MsBackendCached.R
jorainer: Exactly.  This function is a convenience function for extending backends that now don't have to implement the code to return empty (`NA`) core spectra variables.

R/MsBackend.R
```diff
@@ -526,6 +526,11 @@
 #' above). Developers of new `MsBackend`s should follow the
 #' `MsBackendDataFrame` implementation.
 #'
+#' A backend providing a caching mechanism to allow *read only* backends to
+#' add or change spectra variables is the [MsBackendCached()] backend. This
+#' backend is not thought to be used on its own, but backends extending this
+#' backend might be implemented.
```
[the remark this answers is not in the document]
lgatto: See [MsBackendCached()] for details.

R/MsBackendCached.R
```diff
@@ -0,0 +1,614 @@
...
+#' @description
+#'
+#' The `MsBackendCached` class is a rudimentary implementation of the
+#' [MsBackend] providing a simple mechanism to cache spectra data locally. This
+#' class is thought to be used as a base class for other `MsBackend`
+#' implementations to reuse its caching mechanism and avoid having to
+#' re-implement commonly used methods. This class is thus not thought to be
+#' used directly by a user.
+#'
+#' The provided caching mechanism by this class allows `MsBackend` instances
```
lgatto, line 16: The `MsBackendCached` caching mechanism allows ...

R/MsBackendCached.R
```diff
@@ -0,0 +1,614 @@
...
+#' not allow to alter values (e.g. if a SQL database is used as a backend). Any
+#' replacement operation with `$<-` will add the specified values to a local
+#' `data.frame` within the `MsBackendCached` class that allows to *cache* these
+#' values (increasing obviously the memory demand of the object).
+#'
+#' Any data accessor functions of the extending `MsBackend` class (such as `$`
+#' or `msLevel` or `spectraData`) should first use `callNextMethod` to call the
+#' respective accessor of `MsBackendCached` that will evaluate if the
+#' requested spectra variable(s) are in the local cache and return these. If
+#' the requested spectra variables are not in the local cache, are also not in
```
lgatto, line 27: If the requested spectra variables are neither in the local cache, nor listed in the `@spectraVariables` slot

R/MsBackendCached.R
```diff
@@ -0,0 +1,614 @@
...
+#' listed in the `@spectraVariables` slot (which defines all spectra variables
+#' that can be requested from the extending `MsBackend` class) but are *core
+#' spectra variables* then missing values of the correct data type are returned.
+#'
+#' @section Implementation notes:
+#'
+#' Classes extending the `MsBackendCached` need to
+#'
+#' - call the `backendInitialize` method of this class in their own
+#'   `backendInitialize` method and to set at least the number of spectra with
```
lgatto, line 37: One of _and_ or _to_ is too much here.

R/MsBackendCached.R
```diff
@@ -0,0 +1,614 @@
...
+#'   the `nspectra` parameter and the `spectraVariables` that are available to
+#'   the (extending) backend class.
+#'
+#' - implement the `spectraData` method that calls also the `spectraData`
```
lgatto, line 41: that also calls

R/MsBackendCached.R
```diff
@@ -0,0 +1,614 @@
...
+#' implemented by the extending backend class (the default implementations of
+#' the `MsBackendCached` will then be used instead; these ensure that cached
+#' values are returned first).
+#' Spectra variables can be modified or added using the `$<-` method of the
+#' `MsBackendCached`. Replacing or adding multiple variables using the
+#' `spectraData<-` is not supported by `MsBackendCached`. The extending backend
+#' might however implement such a method that internally uses `$<-` to
+#' add/replace single variables.
+#'
+#' The `MsBackendCached` has the following slots"
```
lgatto, line 63: " should probably be :

R/MsBackend.R
```diff
@@ -526,6 +526,11 @@
 #' above). Developers of new `MsBackend`s should follow the
 #' `MsBackendDataFrame` implementation.
 #'
+#' A backend providing a caching mechanism to allow *read only* backends to
+#' add or change spectra variables is the [MsBackendCached()] backend. This
+#' backend is not thought to be used on its own, but backends extending this
```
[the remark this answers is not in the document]
jorainer: Will fix, thanks

R/MsBackend.R
```diff
@@ -526,6 +526,11 @@
...
+#' backend might be implemented.
+#'
```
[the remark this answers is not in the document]
jorainer: Thanks

R/MsBackend.R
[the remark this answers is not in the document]
jorainer: OK

R/MsBackendCached.R
jorainer: Thanks

R/MsBackendCached.R
jorainer: Thanks

R/MsBackendCached.R
jorainer: +1

R/MsBackendCached.R
jorainer: Thanks

R/MsBackendCached.R
jorainer: right

Issue #236, opened 2022-02-22
Create neutralLoss spectra


jorainer:
Add a function `neutralLoss` to create *neutral loss* spectra versions of a `Spectra`. The function could be defined in the following way:

```r
neutralLoss(x, msLevel. = 2L, peaks = c("all", "bigger", "smaller")) {
}
```

With `x` being a `Spectra`, `msLevel.` to define the MS level on which to perform the calculation (MS2 would make most sense obviously) and `peaks` allowing to define in addition if peaks should be pre-filtered:

- `peaks = "all"`: report all fragment peaks.
- `peaks = "bigger"`: only report fragment peaks with an m/z > precursor m/z.
- `peaks = "smaller"`: only report fragment peaks with an m/z < precursor m/z.

Any feedback and suggestions welcome @michaelwitting

jorainer:
I checked the (php) code of https://doi.org/10.1021/jasms.1c00343 (available [here](https://github.com/masspec/MS2ToNLConverter)). Seems they are simply calculating `precursor m/z - fragment m/z` and report the intensity of the fragment peak. No additional filtering is done. Also, no special handling of negative m/z values etc is performed.

jorainer:
Maybe better to have a parameter `filterPeaks = c("none", "largerPmz")` that also supports a user-specified `function` to filter the peaks matrix. `"largerPmz"` would remove all peaks with an m/z that is larger than the one from the precursor m/z (`"Pmz"`).

sgibb:
Maybe you could reuse code from here: https://github.com/lgatto/MSnbase/blob/master/R/functions-fragments.R
(it is already reused/copied to `PSMatch`: https://github.com/rformassspectrometry/PSMatch/blob/main/R/fragments-calculate.R)

jorainer:
Seems that type of neutral loss calculations are different than the one I had in mind (i.e. the neutral loss spectra from https://doi.org/10.1021/jasms.1c00343 that are defined as precursor m/z - fragment m/z and the *hypothetical neutral loss* from https://doi.org/10.1021/acs.analchem.0c02521). 

Considering that, it would maybe be better to define `neutralLoss` as a method for `Spectra` that dispatches on a `Param` object which also defines the type of calculation that should be performed:

- `setMethod("neutralLoss", "Spectra", "NLParam")` to simply calculate precursor m/z - fragment m/z
[...]

PR #237, opened 2022-02-23
feat: add neutralLoss method (issue #236)

jorainer:
- Add `neutralLoss` method.
- Add `PrecursorMzParam` to implement the precursor m/z - fragment m/z neutral
  loss calculation.

Also feedback on naming of the parameter object (`PrecursorMzParam` for the approach to calculate neutral loss spectra by precursor m/z - fragment m/z) welcome.


jorainer:
@michaelwitting, can you please also have a look at this PR to check if this would work for you?

michaelwitting:
Looks good to me so far, need to test with some own data knowning to work with that.

R/Spectra-neutralLoss.R
```diff
@@ -0,0 +1,167 @@
+setClassUnion("functionOrNull", c("function", "NULL"))
+
+#' @title Calculate Neutral Loss Spectra
+#'
+#' @name neutralLoss
+#'
+#' @description
+#'
+#' This help page lists functions to convert MS/MS spectra to neutral loss
```
lgatto, line 9: functions that concert

R/Spectra-neutralLoss.R
```diff
@@ -0,0 +1,167 @@
...
+#' spectra. The main function for this is `neutralLoss` and the specific
+#' algorithm to be used is defined (and configured) with dedicated *parameter*
+#' objects (paramer `param` of the `neutralLoss` function.
```
lgatto, line 12: (argument `param` of the `neutralLoss` function).

R/Spectra-neutralLoss.R
```diff
@@ -0,0 +1,167 @@
...
+#'
+#' The parameter objects for the different algorithms are:
+#'
+#' - `PrecursorMzParam`: calculates neutral loss spectra as in Aisporna 2022 by
```
lgatto, line 16: as in Aisporna *et al.* (2022)

R/Spectra-neutralLoss.R
```diff
@@ -0,0 +1,167 @@
...
+#'   subtracting the (fragment's) peak m/z value from the precursor m/z value
+#'   of each spectrum (precursor m/z - fragment m/z). Parameter `msLevel`
+#'   allows to restrict calculation of neutral loss spectra to specified MS
+#'   level(s). Spectra from other MS level(s) are returned as-is. Parameter
+#'   `filterPeaks` allows in addition to filter fragment peaks e.g. if their
```
lgatto, line 21: Not sure what the 'in addtion' means here. If it's in addition to the other parameters, I would drop it. If it's supposed to mean it has multiple effects, then I'm not sure what they are.

lgatto (changes requested):

R/Spectra-neutralLoss.R
jorainer: You're right. Better drop it.

jorainer:
Changes done @lgatto

jorainer:
Now I'm a little confused about the failing GHA. Locally all runs fine.

jorainer:
OK, unit tests fixed. These were related to the recent change in `MsCoreUtils` (return a positive ppm, even for negative m/z)

lgatto (approved):

PR #240, opened 2022-03-01
Add peaksVariables function

jorainer:
This PR:

- adds the `peaksVariables,MsBackend` method to list the available *peaks variables* from a backend. The default is `"mz"` and `"intensity"`, but backends such as the `MsBackendTimsTof` would also have ion mobility or other peak-related information available.
- change `peaksData,MsBackend` to include a new parameter `columns = peaksVariables()`.

Open for discussion:

- how should we handle the existing `peaksData` implementations? So far I have only added `...` and hence the `columns` parameter would be ignored for the `MsBackendMzR`, `MsBackendDataFrame` et al backends.
- should we change `peaksData,Spectra` to include a parameter `columns = peaksData(object)` or should we support passing that parameter using `...`?


jorainer:
Any opinion on that @lgatto or @sgibb ?

jorainer:
I'll work on that and convert to a PR again once I'm done.

jorainer:
Some general info:
- `peaksData,Spectra` gained parameter `columns` which is passed down to `peaksData,MsBackend`.
- The backends in `Spectra` support only columns `"mz"` and `"intensity"`, but in any arbitrary order.
- This PR will most likely break existing backends because they don't yet support the paramter `columns` - but I will fix these later.

jorainer:
OK for review @sgibb and @lgatto :)

lgatto (approved):

PR #243, opened 2022-04-13
docs: expand examples and description on bin

jorainer:
- Expand documentation and vignette on the usage of the `bin,Spectra` method.
- `bin,Spectra` gains also parameter `FUN` to enable specifying the aggregation
  function.


jorainer:
Merging since it mainly contains expanded documentation related to issue #424.

Issue #246, opened 2022-10-04
A more efficient in-memory backend


jorainer:
Our original `MsBackendDataFrame` backend is not that fast and efficient as originally planned. Using a single `DataFrame` to store all data was most likely not the best idea since a) subsetting of `DataFrame` is slow (compared to a `data.frame` and b) storing all data, also the peaks data within one `DataFrame` has an impact on any (subsetting) operation on that data.

I've now implemented a `MsBackendDF` with this definition:

```r
setClass("MsBackendDF",
         contains = "MsBackend",
         slots = c(spectraData = "data.frame",
                   peaksData = "list",
                   peaksDataFrame = "list"),
         prototype = prototype(spectraData = data.frame(),
                               peaksData = list(),
                               peaksDataFrame = list(),
                               readonly = FALSE,
                               version = "0.1"))
```

Spectra data is thus stored within a `data.frame` (instead of a `DataFrame`) and the peaks data is stored in its more *natural* way, i.e. as a `list` of numerical matrices. Also, I've added a new slot `peaksDataFrame` that would allow to store any additional peaks variable in it (e.g. peak annotations etc).

I did some performance calculations on this backend (see further below) and as expected, it is faster than the original `MsBackendDataFrame`. What I would now suggest is to simply **add** this backend as an additional backend to the `Spectra` package (still keeping the `MsBackendDataFrame` because other backends extend that) and change the `MsBackendDataFrame()` constructor to return a `MsBackendDF` instead of a `MsBackendDataFrame`. I think that should be a minimally invasive change that would switch the default in-memory backend from `MsBackendDataFrame` to `MsBackendDF`.

Any thoughts @lgatto and @sgibb ? If you're OK with it I would then go ahead and make a PR.

jorainer:
The promised performance comparisons:

```r
library(Spectra)
library(msdata)

fl <- system.file(
    "proteomics",
    "TMT_Erwinia_1uLSike_Top10HCD_isol2_45stepped_60min_01-20141210.mzML.gz",
    package = "msdata")
sps_od <- Spectra(fl)
sps_im <- setBackend(sps_od, backend = MsBackendDataFrame())
sps_df <- setBackend(sps_od, backend = new("MsBackendDF"))
```

Access of a single spectra variable:

```r
library(microbenchmark)
microbenchmark(msLevel(sps_im),
               msLevel(sps_df))
Unit: microseconds
##            expr     min       lq     mean   median       uq     max neval cld
[...]

PR #247, opened 2022-10-05
Add new default in memory backend MsBackendMemory

jorainer:
This PR adds the new `MsBackendMemory` in-memory backend discussed in #246 : 
- suggestions from @lgatto have been considered.
- documentation and vignette was updated to use by default the `MsBackendMemory`.


jorainer:
@lgatto or @sgibb , would be nice if one of you could have a look at it so that I could merge it **before** the BioC release deadline.

R/MsBackendMemory-functions.R
```diff
@@ -0,0 +1,110 @@
...
+        }
+        if (any(columns == "mz"))
+            res$mz <- mz(object)
+        if (any(columns == "intensity"))
+            res$intensity <- intensity(object)
+        p_columns <- setdiff(p_vars, c("mz", "intensity"))
+        for (p_column in p_columns) {
+            res <- do.call(
+                `$<-`, list(res, name = p_column,
+                            value = .df_pdata_column(object@peaksDataFrame,
```
sgibb, line 60: Wouldn't it be possible to return multiple columns via `.df_pdata_column` to avoid this `for` loop?
(Maybe if we just drop the `which` statement in the first line of `.df_pdata_column` or replace it by `%in%`?)

vignettes/Spectra.Rmd
```diff
@@ -250,15 +250,40 @@ pks <- peaksData(sps)
...
+variable.
+
+```{r}
+## Peak annotations for the first spectrum
+peaksData(sps2, "peak_anno")[[1L]]
+## Peak annotations for the second spectrum
+peaksData(sps2, "peak_anno")[[2L]]
+```
+
+This works however **only** for `Spectra` that use a `MsBackendMemory` backend.
```
sgibb, line 285: Couldn't this be implemented by all writeable backends? (but is currently just supported by `MsBackendMemory`?

sgibb (changes requested):
Sorry for the delayed review. However all in all a nice PR. I have just a few small questions.

lgatto (commented):
Here's a first round of comments/questions. I haven't looked at the code yet, but wanted to first agree on some high level points.

R/MsBackend.R
```diff
@@ -475,14 +479,33 @@
...
-#' New objects can be created with the `MsBackendDataFrame()`
-#' function. The backend can be subsequently initialized with the
-#' `backendInitialize` method, taking a `DataFrame` with the MS data
-#' as parameter. Suggested columns of this `DataFrame` are:
+#' @section In-memory data backends: `MsBackendMemory` and `MsBackendDataFrame`:
+#'
+#' The `MsBackendMemory` and `MsBackendDataFrame` objects keep all MS data in
+#' memory are thus ideal for fast data processing. Due to their large memory
+#' footprint they are however not suited for large scale experiments. The two
+#' backends store the data internally different. The `MsBackendDataFrame` stores
```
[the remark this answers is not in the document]
jorainer: Agree, thanks

R/MsBackend.R
```diff
@@ -475,14 +479,33 @@
...
+#' full peak data (i.e. the numerical matrices of m/z and intensity values).
+#' Also subsetting and access to any spectra variable (except `"mz"` and
+#' `"intensity"` is fastest for the `MsBackendMemory`. Finally, the
+#' `MsBackendMemory` supports also arbitrary peak annotations while the
+#' `MsBackendDataFrame` does not have support for such additional peak
+#' variables.
+#'
+#' Thus, for most use cases, the `MsBackendMemory` provides a higher
+#' performance and flexibility than the `MsBackendDataFrame` and should thus be
+#' preferred.
```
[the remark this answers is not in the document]
jorainer: @sgibb : I would also not deprecate the `MsBackendDataFrame` - might be useful for other use cases (like what Laurent suggested).

@lgatto : I would like to test additional peak annotations first. We don't have yet any real use cases for them, but as said, I'm working on that, and once we know how to deal with them we could add them also to `MsBackendDataFrame`.
And yes, I will add a link to the issue in the documentation.

Regarding `joinSpectraData`: I need to test that. Will report back on that.

R/MsBackendMemory-functions.R
```diff
@@ -0,0 +1,110 @@
...
+#' @noRd
+.df_peaks_columns_data_frame <- function(x) {
+    lns <- integer()
+    if (any(colnames(x) == "mz"))
+        lns <- lengths(x$mz)
+    if (!length(lns) && any(colnames(x) == "intensity"))
+        lns <- lengths(x$intensity)
+    if (length(lns)) {
+        colnames(x)[vapply(x, function(z) {
+            (inherits(z, "NumericList") || inherits(z, "SimpleList") ||
```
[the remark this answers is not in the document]
jorainer: I would say `list` would be the most likely. I will change, thanks.

R/MsBackendMemory-functions.R
jorainer: I would not expect a major performance gain. But PR would be welcome ;)

R/Spectra.R
```diff
@@ -52,7 +53,7 @@ NULL
 #' which supports the following formats:
 #'
 #' - parameter `object` is a `DataFrame` containing the spectrum data. The
-#'   provided `backend` (by default a [MsBackendDataFrame-class]) will be
+#'   provided `backend` (by default a [MsBackendMemory-class]) will be
```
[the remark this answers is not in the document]
jorainer: `MsBackendMemory` is now the default for `Spectra,ANY`. But `object` can now be either a `data.frame` OR a `DataFrame`. I will update the documentation.

R/MsBackendMemory.R
```diff
@@ -0,0 +1,657 @@
...
+#'
+#' @importFrom methods as
+#'
+#' @importFrom S4Vectors SimpleList
+#'
+#' @importMethodsFrom S4Vectors lapply
+setMethod(
+    "spectraData", "MsBackendMemory",
+    function(object, columns = spectraVariables(object)) {
+        as(.df_spectra_data(object, columns), "DataFrame")
```
sgibb, line 536: Does this have to return a `DataFrame`? (is this needed for compatibility reasons?)

R/MsBackendMemory.R
```diff
@@ -0,0 +1,657 @@
...
+    })
+
+#' @rdname hidden_aliases
+#'
+#' @note: this replaces **all** the data in the backend.
+setReplaceMethod("spectraData", "MsBackendMemory", function(object, value) {
+    if (inherits(value, "DataFrame")) {
```
sgibb, line 543: Shouldn't we allow a `data.frame` here as well?

Looks wired:

```r
d <- data.frame(a = 1:3)
# internally it is converted in a data.frame
spectraData(msbckmem) <- as(d, "DataFrame")
# internally it is converted from a data.frame to a DataFrame
all.equal(as.data.frame(spectraData(msbckmem)), d)
```

sgibb (changes requested):
Somehow I completely missed the class definition. The file was hidden because of the large number of changes.

jorainer:
@lgatto , I had a look at `joinSpectraData`. The function works - until it comes to add `NumericList` and other `S4` objects. Maybe best if I add a check to the `$<-,MsBackendMemory` method to test whether `value` is an `S4` object and if so throw an error with the suggestion to change to backend to `MsBackendDataFrame`.

R/MsBackendMemory.R
```diff
@@ -0,0 +1,657 @@
...
+#' @importClassesFrom IRanges NumericList
+#'
+#' @importFrom IRanges NumericList
+#'
+#' @rdname MsBackend
+setMethod("backendInitialize", signature = "MsBackendMemory",
+          function(object, data, ...) {
+              if (missing(data)) data <- data.frame()
+              if (is(data, "DataFrame"))
+                  data <- as(data, "data.frame")
```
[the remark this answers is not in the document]
jorainer: I would avoid that: `setBackend` calls also this function, and will submit all the data from the *old* backend as a `DataFrame` to the new backend. Thus, the user would always get this warning whenever he/she changes to an `MsBackendMemory` backend.

But like I mentioned to Laurent below (above?), I think we should add a check to the `<-,MsBackendMemory` to test whether `value` is an `S4` object - because storing them in the (internal) `data.frame` could eventually not work.

R/MsBackendMemory.R
jorainer: exactly. That's our convention.

jorainer: Back from the days where we were not yet aware of the performance difference...

R/MsBackendMemory.R
jorainer: yes, you're right. Maybe best to support `data.frame` and `DataFrame`.

R/MsBackendMemory.R
lgatto: But do we want to keep this? Wouldn't it make more sense to return the slot as it is?

lgatto:
As mentioned above for `spectraData()`, I would suggest for `joinSpectraData(x, y)` to return the same class as `spectraData(x)`. It would be the user's responsibility to use whatever matches their needs.

R/MsBackendMemory.R
jorainer: I think it's cleaner that way. We defined in the `MsBackend` specification that the `spectraData` returns a `DataFrame` and that `spectraData<-` expects a `DataFrame`. How the data is stored internally is up to the backend, but we have to be consistent what each backend returns or takes. I think it's important to define a clean and consistent interface. Otherwise we could get into trouble in `Spectra`, needing to check what is returned by the backend, reformatting etc.

jorainer:
@sgibb @lgatto , I think we have to be careful here with the discussion on `DataFrame` *vs* `data.frame`. I would **not** want to change the definition of our `MsBackend` interface to allow backends to return `data.frame` instead of `DataFrame` with e.g. the `spectraData` method. Let's be consistent here and take the object that covers more use cases (e.g. possibility to store `S4` objects as spectra variables).

For the actual `MsBackend` implementations, internally they are allowed to store the data in different ways. Here some things will not be possible, like storing `S4` classes as spectra variables in e.g. an SQL database or in an HDF5 array or, like in this case here, in a `data.frame`. But that's OK since we have alternative backends that support these things - we just have to show an error or warning to the user telling that something is not possible with that backend. 
For the `MsBackendMemory` I will add internal checks to test whether users provide `S4` classes as spectra variables. I will not change the `spectraData,MsBackendMemory` and `spectraData<-,MsBackendMemory` methods to support `data.frame` in addition to `DataFrame` - they have to be consistent and compliant with the `MsBackend` definition. The exception is the `backendInitialize` method which is defined in a very flexible way and accepts any type of input as long as the backend understands how to deal with it.

I will convert this now to a draft and re-invite you for review once I'm done with some more checks.

R/MsBackendMemory.R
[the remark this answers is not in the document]
jorainer: Update: seems we can even have `S4` objects in columns of a `data.frame`. Tested with `List`, `NumericList`.

R/MsBackendMemory.R
jorainer: thinking it over, no, I don't think it's a good idea. This would break the convention of the `MsBackend` interface. All backends return a `DataFrame` with `spectraData` and are supposed to take a `DataFrame` as input. I would not want to change this now for this one backend.

jorainer:
I've updated the `$<-,MsBackendMemory` method. With that we can also assign `list`, `SimpleList` et al as spectra variables (before it would only allow that for peaks variables). With that change now also `joinSpectraData` works as expected.

PR would be ready for review. Please let me know if you don't have the time to review - I would really like to get this in before the Bioc release freeze.

PR #254, opened 2022-12-06
fix: MsBackendMzR of lenght 0

jorainer:
- Fix error extracting spectra data from an `MsBackendMzR` of length 0.


jorainer:
Merging this bug fix.

PR #255, opened 2022-12-14
refactor: use dplyr::between instead of MsCoreUtils::between

jorainer:
Use `dplyr::between` instead of `MsCoreUtils::between`. See also issue in `MsCoreUtils` [here](https://github.com/rformassspectrometry/MsCoreUtils/issues/105) for a performance comparison.


sgibb (changes requested):
Sorry, I would reject this PR in its current form. Please see my comment at https://github.com/rformassspectrometry/MsCoreUtils/issues/105#issuecomment-1350730506.

jorainer:
Excellent. Yes, then let's drop this.

Issue #258, opened 2022-12-25
Bug in containsMz()


lgatto:
## Test data
```
> sciex_file <- dir(system.file("sciex", package = "msdata"),
+ full.names = TRUE)
> sciex <- Spectra(sciex_file, backend = MsBackendMzR())
```

## BUG: mz must be ordered

```
> containsMz(sciex, mz = c(123.7, 109.07), which = "any")
Error: BiocParallel errors
  2 remote errors, element index: 1, 2
  0 unevaluated and other errors
  first remote error:
Error in closest(x, table, tolerance = tolerance, ppm = ppm, duplicates = duplicates, : 'x' and 'table' have to be sorted non-decreasingly and must not contain NA.
> containsMz(sciex, mz = c(109.07, 123.7), which = "any")
  [1] TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE
 [16] TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE
 [31] TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE
 [46] TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE
 [61] TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE
 [76] TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE
 [91] TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE
 [ reached getOption("max.print") -- omitted 1762 entries ]
```

## Suggested fix:

Add `mz <- sort(mz)` at [the beginning of the function](https://github.com/rformassspectrometry/Spectra/blob/master/R/Spectra.R#L1567-L1568), possibly with a message indicating that the mz values get sorted. 

@jorainer @sgibb - I'll send a PR. Any comments/suggestion?

lgatto:
Here's the suggested addition:
```
    if (!identical(order(mz), seq_along(mz))) {
        message("Ordering mz values.")
        mz <- sort(mz)
    }
[...]

PR #259, opened 2022-12-25
fix: bug in containsMz() when mz is not sorted (see #258)

lgatto:


R/Spectra.R
```diff
@@ -1565,6 +1565,11 @@ setMethod("containsMz", "Spectra", function(object, mz = numeric(),
                                             ppm = 20, which = c("any", "all"),
                                             BPPARAM = bpparam()) {
     cond_fun <- match.fun(match.arg(which))
+    if (!identical(order(mz), seq_along(mz))) {
```
sgibb, line 1568: This should be `is.unsorted`.

```suggestion
    if (is.unsorted(mz)) {
```

``` r
library("microbenchmark")

f1 <- function(x)(!identical(order(x), seq_along(x)))
f2 <- function(x)(is.unsorted(x))

n <- c(1, 2, 5, 100, 1000)

set.seed(123)
l <- lapply(n, function(nn) {
    x <- sample(nn)
    message("unsorted")
    print(microbenchmark(f1(x), f2(x), check = "identical"))
    message("sorted")
    x <- sort(x)
    print(microbenchmark(f1(x), f2(x), check = "identical"))
    NULL
})
#> unsorted
#> Unit: microseconds
#>   expr   min     lq      mean median     uq       max neval cld
#>  f1(x) 3.371 3.4800 295.58717 3.5650 3.6685 29180.274   100   a
#>  f2(x) 1.272 1.3185  19.24412 1.4475 1.5505  1775.806   100   a
#> sorted
#> Unit: microseconds
#>   expr   min     lq    mean median     uq    max neval cld
#>  f1(x) 3.602 3.7520 4.16168 3.8340 3.9150 19.213   100   b
#>  f2(x) 1.434 1.5315 1.70864 1.6025 1.7145  7.877   100  a
#> unsorted
#> Unit: microseconds
#>   expr   min     lq    mean median     uq    max neval cld
#>  f1(x) 4.049 4.2645 5.92054  4.712 5.8730 22.581   100   b
#>  f2(x) 2.054 2.2115 3.49219  2.465 3.1925 28.604   100  a
#> sorted
#> Unit: microseconds
#>   expr   min     lq    mean median    uq    max neval cld
#>  f1(x) 3.933 4.1105 4.41991 4.2175 4.302 20.929   100   b
#>  f2(x) 2.032 2.1285 2.53172 2.2430 2.396 27.760   100  a
#> unsorted
#> Unit: microseconds
#>   expr    min      lq     mean  median      uq    max neval cld
#>  f1(x) 29.946 30.6690 33.36808 31.1255 32.6365 92.298   100   b
#>  f2(x)  1.930  2.1565  2.74725  2.5675  2.8485 13.486   100  a
#> sorted
#> Unit: microseconds
#>   expr   min     lq    mean median     uq    max neval cld
#>  f1(x) 3.815 3.9135 4.38598 4.0045 4.1925 15.425   100   b
#>  f2(x) 1.921 2.0450 2.29715 2.1380 2.2305 10.565   100  a
#> unsorted
#> Unit: microseconds
#>   expr   min      lq     mean  median      uq     max neval cld
#>  f1(x) 31.44 32.2060 35.73402 32.5655 32.9870 114.536   100   b
#>  f2(x)  2.07  2.2245  2.79570  2.6090  2.8945  13.987   100  a
#> sorted
#> Unit: microseconds
#>   expr   min    lq    mean median     uq    max neval cld
#>  f1(x) 4.283 4.589 5.99422  4.687 4.8225 69.879   100   b
#>  f2(x) 1.914 1.999 2.24910  2.101 2.2215 12.940   100  a
#> unsorted
#> Unit: microseconds
#>   expr    min      lq     mean  median      uq     max neval cld
#>  f1(x) 39.801 41.7595 49.83098 42.4825 43.0685 340.579   100   b
#>  f2(x)  3.177  3.4400  4.45858  3.8770  4.2925  25.576   100  a
#> sorted
#> Unit: microseconds
#>   expr   min     lq    mean median     uq    max neval cld
#>  f1(x) 5.841 6.0765 7.38256 6.2290 6.5415 74.087   100   b
#>  f2(x) 1.900 2.0310 2.19436 2.1305 2.2230  7.385   100  a
```

<sup>Created on 2022-12-25 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/Spectra.R
```diff
@@ -1565,6 +1565,11 @@ setMethod("containsMz", "Spectra", function(object, mz = numeric(),
...
+        message("Ordering mz values.")
```
sgibb, line 1569: Do we need this message at all? E.g. in `MsCoreUtils::between` we silently reorder the `range` argument if it is unsorted. If we want to keep this message we could add a `verbose` or `quiet` argument (in `MSnbase` we have `Msnbase::isMsnbaseVerbose()`, do we need something similar in `Spectra`?).

sgibb (changes requested):

R/Spectra.R
lgatto: Thanks you! I look `is.sorted()` up and didn't think to look for `is.unsorted()`!

R/Spectra.R
lgatto: I don't mind. It's easy enough to silence with suppressMessage(). The reason I put the message in is a bit more complex (see #260).  

Let's drop it here.

jorainer:
I'm wondering if this check should not go into the (backend) validator methods? That should/could be more efficient since the check would only called once instead of each time the function e.g. `containsMz` is called. AFAIK we require the m/z values to be sorted. The `MsBackendDataFrame` has this check (https://github.com/rformassspectrometry/Spectra/blob/master/R/MsBackendDataFrame-functions.R#L55:L63) - maybe we should add a  validator method for the `MsBackend` class and add the sorted m/z check there?

lgatto:
The last commit implements what what discussed in #258

jorainer (approved):
Thanks for the PR Laurent - and sorry for my confusion. Just one minor thing: does this change now also work if one of the input `mz` values is `NA`?

lgatto:
```
> sciex_file <- dir(system.file("sciex", package = "msdata"),
+                   full.names = TRUE)
+ sciex <- Spectra(sciex_file, backend = MsBackendMzR())
> x1 <- containsMz(sciex, mz = c(123.7, NA, 109.07), which = "any")
> x2 <- containsMz(sciex, mz = c(109.07, 123.7, NA), which = "any")
> x3 <- containsMz(sciex, mz = c(109.07, 123.7), which = "any")
> identical(x1, x2)
[1] TRUE
> identical(x1, x3)
[1] TRUE
```

jorainer (approved):
Thanks Laurent! I will merge and push to BioC

PR #265, opened 2023-01-12
Add first content to the MsBackend howto

jorainer:
Add some first content to the vignette describing how to implement a `MsBackend` class.


jorainer:
@meowcat @Adafede , the vignette currently describes all functions to create a *read-only* `MsBackend`. I'm currently adding information on the replacement menthods.

jorainer (commented):
Will merge that.

Issue #269, opened 2023-02-01
Re-evaluate automatic detection of peaks variables


jorainer:
The `MsBackendMemory` supports additional, arbitrary peaks variables which are stored separately from spectra variables. The `$<-` method tries to automatically detect such peaks variables and assigns them. This should be changed to let the user specifically select or define whether the variable should be a peaks variable or not.

jorainer:
To be consistent with all other backends it's better to disable the automatic detection/conversion of spectra variables to peaks variables in `MsBackendMemory` as that can result in unexpected behavior.

Issue #268, opened 2023-01-24
`setBackend` to `MsBackendMemory` fails for empty `Spectra`


meowcat:
`setBackend(sp, MsBackendMemory())`` fails for empty `sp`:
```R
fls <- dir(system.file("sciex", package = "msdata"), full.names = TRUE)
sp <- Spectra(fls, source = MsBackendMzR())
sp
# MSn data (Spectra) with 1862 spectra in a MsBackendMzR backend:
sp_zero <- sp[c()]
sp_zero
# MSn data (Spectra) with 0 spectra in a MsBackendMzR backend:
sp_zero_mem <- setBackend(sp_zero, MsBackendMemory())
# Error in validObject(object) : 
# invalid class “MsBackendMemory” object: The following columns have a wrong data type: mz, intensity. 
# The expected data type(s) is/are: NumericList, NumericList.
```

This is not specific to `MsBackendMzR`, but also happens with `MsBackendDataFrame`:

```R
sp_df <- setBackend(sp, MsBackendDataFrame())
sp_df_zero <- sp_df[c()]
sp_df_zero_mem <- setBackend(sp_df_zero, MsBackendMemory())
# Error in validObject(object) : 
# invalid class “MsBackendMemory” object: The following columns have a wrong data type: mz, intensity. 
# The expected data type(s) is/are: NumericList, NumericList.
```

On the other hand, converting an empty `Spectra` to `MsBackendDataFrame` works fine, therefore I believe (and I think) this is the expected behaviour, and `MsBackendMemory` is the "wrong" one:
```R
sp_zero <- sp[c()]
sp_zero_df <- setBackend(sp_zero, MsBackendDataFrame())
sp_zero_df 
# MSn data (Spectra) with 0 spectra in a MsBackendDataFrame backend:
# Processing:
# Switch backend from MsBackendMzR to MsBackendDataFrame [Tue Jan 24 13:50:08 2023] 
```

jorainer:
Thanks for pointing this out - I will have a look into it and fix. I believe the standard behavior for `setBackend` should be to allow empty  spectra too.

jorainer:
this should be fixed with the PR above.

Issue #267, opened 2023-01-24
MzBackendMzR misleading error with empty file list and misleading behavior with `files` argument


meowcat:
When constructing `MzBackendMzR` with an empty file list, the returned error message is misleading:

```R
fls <- c()
sp <- Spectra(fls, source = MsBackendMzR())
# Error in .local(object, ...) : 
#   Parameter 'files' is mandatory for 'MsBackendMzR'
```

This lets me think that I should be using `Spectra(files = fls, ...)` instead, which in fact does not error:
```R
sp <- Spectra(files = fls, source = MsBackendMzR())
# no error
sp
# MSn data (Spectra) with 0 spectra in a MsBackendMemory backend:
```

However `files` is not the right argument, in fact `sp <- Spectra(files = fls, source = MsBackendMzR())` completely ignores `fls`:
```R
fls <- dir(system.file("sciex", package = "msdata"), full.names = TRUE)
sp <- Spectra(files = fls, source = MsBackendMzR())
sp
# MSn data (Spectra) with 0 spectra in a MsBackendMemory backend:
```

Passing `fls` as anonymous parameter works as expected:
```R
sp <- Spectra(fls, source = MsBackendMzR())
sp
# MSn data (Spectra) with 1862 spectra in a MsBackendMzR backend:
```

**Expected behaviour:**
`fls <- c(); sp <- Spectra(fls, source = MsBackendMzR())` should 
* either return a valid `Spectra` object with 0 spectra and potentially warn that the file list was empty
* or error because the file list was empty.

Note that no-input `MsBackendMzR` (and also `MsBackendMgf`) objects seem to be invalid anyway. I don't know if this is intended, so this would influence what behaviour we expect from a no-file `MzBackendMzR`.
```R
sp <- Spectra(source = MsBackendMzR())
sp2 <- setBackend(sp, MsBackendMemory())
# Error in validObject(object) : 
# invalid class “MsBackendMemory” object: The following columns have a wrong data type: mz, intensity.
# The expected data type(s) is/are: NumericList, NumericList
```

(Discovered accidentally because I did not have package "msdata" installed when building the `Spectra.Rmd` vignette.)

meowcat:
I spun out the last part into a separate issue #268

jorainer:
Good point Michele. Will fix.

jorainer:
this should be fixed with the PR above.

PR #270, opened 2023-02-02
Fix issue with empty MsBackendMemory and automatic peak variable detection in MsBackendMemory

jorainer:
This PR addresses issues #268 and #269:

- #268: fix support for empty `Spectra` and `MsBackendMemory`.
- #269: change the original way how peaks variables are handled in `MsBackendMemory`: instead of guessing if a spectra variable is in fact a peaks variable (based on the data type and number of elements), additional peaks variables must now be specified by the user using the `peaksVariables` parameter in `backendInitialize`. This ensures compliance of `MsBackendMemory` with all other backends but still provides the user the possibility to use the additional functionality from `MsBackendMemory`.


jorainer:
Just added also a fix for issue #267

R/MsBackend.R
```diff
@@ -720,18 +744,32 @@
 #' ## The `MsBackendMemory` uses a more efficient internal data organization
 #' ## and allows also adding arbitrary additional peaks variables (annotations)
 #' ## Below we thus add a column "peak_ann" with arbitrary names/ids for each
-#' ## peak
+#' ## peak and add the name of this columnt to the `peaksVariables` parameter
```
sgibb, line 747:
```suggestion
#' ## peak and add the name of this column to the `peaksVariables` parameter
```

sgibb (approved):
IMHO the implementation is fine but I didn't look at the unit test.

Issue #271, opened 2023-02-06
Check if a backend supports parallel processing


jorainer:
This is to keep track of the issue from @meowcat (https://github.com/rformassspectrometry/Spectra/issues/262#issuecomment-1403846824):

Some backends, specifically those using SQL databases don't support parallel processing (because the SQL connection can not be shared across threads). We need to consider such situations since we can get unexpected errors on some backends.

What seems the best option is to add a method `backendBpparam` to `MsBackend` with the following default:

```r
setMethod("backendBpparam", "MsBackend", function(object, BPPARAM = bpparam()) {
    BPPARAM
})
```

this should be called by all `Spectra` methods that run any `BiocParallel` method to validate/check the user defined `BPPARAM` parameter. A `MsBackendSql` backend would thus implement:

```r
setMethod("backendBpparam", "MsBackendSql", function(object, BPPARAM = bpparam()) {
    if (!inherits(BPPARAM, "SerialParam"))
        message("Disabled parallel processing since 'MsBackendSql' does not support that")
    SerialParam()
})
```

meowcat:
Copying from the other issue:

> Even database connections would work, if the backend opens and closes the connection for each process.

Do you know of any way to force a side effect when opening a new process?

In my case, I could in theory `if(!dbIsValid(db@con)) db@con <- dbConnect(...)` in every function (or better, get `con` via an accessor function rather than from the slot). However this would reconnect on every use of the Spectra object until it is copied / subsetted / filtered, because the resulting new object with a working `con` is ephemeral. Not sure if I am being clear.

(A way around it would be to additionally maintain a connection table in some global object in the background.)

meowcat:
Storing the `con` in an `environment` slot should work.

jorainer:
From my old days of java programming - what we did there is to initialize a pool of database connections and processes could get one from there - or a new one is created if none is available. what I don't quite like about that idea is that we would need to store the connection information within the backend. The question is really if there is any benefit from parallel database queries? some databases might not even allow that.

Regarding storing the `con` in an `environment` - are you sure about that? I guess that would not work on Windows because there a new R process is started for each parallel process and no memory is shared across them (thus I guess even there the `environment` would need to be copied - but I didn't try...).

I haven't tried the benefits/disadvantages of keeping the connection in the backend with opening/closing the connection for each call. Both approaches should work. and the latter would support parallel processing...

meowcat:
[...]

PR #272, opened 2023-02-17
Add backendBpparam method and minor tweaks

jorainer:
- Add the `backendBpparam` method allowing to evaluate whether a `MsBackend` supports parallel processing.
- Use `backendBpparam` across `Spectra` methods to eventually disable parallel processing for backends not supporting it.
- Minor tweaks in `.peaksapply` and spectra comparison functions: avoid unnecessary splitting/joining of data if no parallel processing is performed and avoid repeated function calls in loop.

This addresses issue #271 avoiding unexpected errors with `Spectra` objects using e.g. database backends.


jorainer (commented):
Had another (fresh) look at the code and to me all looks good. I will merge in the next couple of days if no other review comes in.

andreavicini (approved):
All looks good to me! :)

PR #273, opened 2023-03-02
 Add MsBackendRawFileReader  in backend listing

cpanse:


vignettes/Spectra.Rmd
```diff
@@ -1017,6 +1017,11 @@ could be used instead. The default backends are:
 - `MsBackendMzR`: this backend keeps only general spectra variables in memory
   and relies on the `r Biocpkg("mzR")` package to read mass peaks (m/z and
   intensity values) from the original MS files on-demand.
+  
+- `r Biocpkg("MsBackendRawFileReader")`: implements a backend for reading
```
jorainer, line 1021: I would suggest to move this paragraph a bit down were other (external) packages providing backends are listed (line 1044 onwards where e.g. `MsBackendMgf` etc are listed).

Also, maybe I would remove the "Orbitrap" and clearly speficy that your backend support importing data directly from the **raw** data files for Thermo MS Instruments.

jorainer (changes requested):
Thanks Christian for the PR. I would suggest some minor changes.

vignettes/Spectra.Rmd
cpanse: yes; fine with me. only one thing. can you replace `Thermo` with `Thermo Fisher Scientific`.

jorainer (approved):
Thanks Christian. I will merge now.

PR #276, opened 2023-03-31
fix: selectSpectraVariables,MsBackendMemory

jorainer:
- Fix `selectSpectraVariables,MsBackendMemory`: return values for *core* spectra variables `"mz"` and `"intensity"` even if they are not in peaks data.


jorainer:
Merging because contains a simple fix.

PR #277, opened 2023-04-07
Fix filter precursor scan

lgatto:
This PR fixes issue #194 

This fix is rather simple - after running `filterPrecursorScan(.)` on each file (as was implemented by @jorainer in https://github.com/rformassspectrometry/Spectra/commit/12021bedd92206fbca5388cc009a8b63291302e9, I added a `filterDataOrigin(., f)`.


jorainer (approved):
Looks good to me. Thanks @lgatto

Issue #289, opened 2023-05-22
API CHANGE: peaksData to return data.frame instead of matrix


jorainer:
This is related to issue #287 and was discussed [here](https://github.com/rformassspectrometry/RforMassSpectrometry/issues/32): change the `MsBackend` API to support `peaksData` to return a `list` of `data.frame` instead (or in addition) to a `list` of `matrix`. This would enable *native* support of peak variables (annotations) that are not `numeric`.

This change *should* not break any functionality, but we need to check also if and how strong it's impact on the performance is.

jorainer:
Comparing impact of returning a `data.frame` instead of `matrix` by `peaksData`:

```r
library(Spectra)
fl <- system.file("TripleTOF-SWATH", "PestMix1_DDA.mzML",
                  package = "msdata")
be <- backendInitialize(MsBackendMzR(), fl)

mem <- backendInitialize(MsBackendMemory(), spectraData(be))
mem2 <- backendInitialize(MsBackendMemory2(), spectraData(be))

class(mem@peaksData[[1L]])
[1] "matrix" "array" 
class(mem2@peaksData[[1L]])
[1] "data.frame"
```

The `MsBackendMemory2` is essentially identical to the `MsBackendMemory` only that it keeps `"mz"` and `"intensity"` values in a `data.frame` instead of a `matrix`.

## Tests at the `MsBackend` level

Performance of accessing `peaksData`, `mz` and `intensity`.

```r
library(testthat)
library(microbenchmark)
microbenchmark(peaksData(mem), peaksData(mem2))
Unit: microseconds
            expr   min     lq    mean median     uq    max neval cld
  peaksData(mem) 5.486 5.6200 5.81063 5.6945 5.8110 13.633   100   a
 peaksData(mem2) 5.663 5.7825 6.25971 5.8550 5.9495 42.090   100   a
```

No difference (not unexpectedly).

```r
expect_equal(mz(mem), mz(mem2))
microbenchmark(mz(mem), mz(mem2), times = 13)
Unit: milliseconds
     expr      min       lq     mean   median       uq      max neval cld
  mz(mem) 36.67263 38.38423 40.14132 38.97847 41.48369 49.28339    13  a 
 mz(mem2) 46.58165 48.36403 51.12181 51.03753 52.70381 58.58926    13   b
```

The `matrix`-based backend is about 1.3 times faster. Note: access using `$mz` and `$intensity` instead of `[[` would increase performance again

```r
library(IRanges)
[...]

Issue #287, opened 2023-05-17
Add support for peak annotations


jorainer:
`Spectra` supports adding/changing *spectra variables* but there is only limited support for *peaks variables*. Some backends (SQL-based backends) support extracting additional peaks variables/annotations from the database and provide these as additional columns in `peaksData`. Also, the `MsBackendMemory` supports additional peaks variables that are stored internally in a `data.frame`.

Since `peaksData` returns a `list` of `matrix`, adding character peak annotations is not possible. Thus, one suggestion would be to have an (additional) `peaksDataFrame` function that returns the data for requested peaks variables as a `data.frame` and `peaksDataFrame<-` would support changing/updating/adding peaks variables. Alternatively, we could have a `peaksAnnotation` function that returns peak annotations (everything except the m/z and intensity values).

Note: support for peak annotations/variables will also need some way to ensure data does not get corrupted when peak matrices are filtered with functions such as `filterMzValues` or `filterMzRange` that reduce the number of peaks.

Issue #288, opened 2023-05-18
Split Spectra documentation into smaller chunks


jorainer:
Instead of having all documentation for `Spectra` in one (large) man page, split them by topic:

- General `Spectra` documentation including creation.
- Subsetting and filtering of `Spectra`.
- Data manipulation.
- Plotting.

PR #297, opened 2023-06-15
Add peak variable support to MsBackendMemory and MsBackendDataFrame

jorainer:
This PR adds the peak variable functionality mentioned in issues #289 (and #287). In essence the *promise* of `peaksData` is to return a 2 dimensional data structure. Can be a `matrix` or `data.frame`. If only `"mz"` and `"intensity"` are requested `MsBackendMemory` and `MsBackendDataFrame` return a `list` of `matrix`, otherwise a `list` of `data.frame`.


lgatto (changes requested):

NEWS.md
```diff
@@ -19,6 +19,23 @@
...
+
+- Import `uniqueMsLevels` from `ProtGenerics`.
+
+## Changes in 1.11.3
+
+- Rename `combinePeaks` for lists of peak matrices into `combinePeaksData`.
+- Add `combinePeaks` generics.
+- Add `combinePeaks,Spectra` to combine peaks within each spectrum in a
+  `Spectra`.
+>>>>>>> main
```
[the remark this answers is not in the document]
jorainer: fixed now :)

R/Spectra.R
```diff
@@ -1887,11 +1927,27 @@ setReplaceMethod("smoothed", "Spectra", function(object, value) {
...
 #' @exportMethod spectraData
-setMethod("spectraData", "Spectra", function(object,
-                                             columns = spectraVariables(object))
-{
-    spectraData(object@backend, columns = columns)
-})
+setMethod(
+    "spectraData", "Spectra",
+    function(object, columns = spectraVariables(object)) {
+        if (length(object@processingQueue) &&
```
lgatto, line 1933: Just a suggestion, but I would find it useful to have some comments that describe the different situations here.

jorainer: you mean within the code?

jorainer: I added comments in the code. Basically, if the processing queue is empty or the user does **not** request peak variables with `columns` the `spectraData` method from the backend is called. Otherwise we need to execute the processing queue (calling `peaksData` on the `Spectra` object) to ensure that e.g. peaks are filtered etc and add the individual peaks variables to the returned `DataFrame`.

jorainer:
> For now, additional peaks variables are only possible for in-memory (i.e. Memory and DataFrame) backends, right?

Actually, no. Since e.g. `MsBackendMzR` and others inherit from `MsBackendDataFrame` (or `MsBackendMemory`) they support also peaks variables. For `MsBackendMzR` they would obviously not (yet) imported from mzML files directly, but it would be possible to add them later (with a future `addPeaksVariable` function). There is already some notes on peaks variable support of different backends in the (too large) documentation of `Spectra` objects. But that could then also be improved in the revamp of the docs (issue #288 )

R/MsBackendDataFrame-functions.R
```diff
@@ -69,6 +69,26 @@ NULL
     NULL
 }
 
+.peaks_variables <- function(x) {
+    if (x@version != "0.1") {
```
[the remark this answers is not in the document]
jorainer: fixed now

jorainer:
Thanks @lgatto for the review. I have now addressed the comments.

R/MsBackend.R
```diff
@@ -544,10 +546,7 @@
 #' way the data is organized internally, provides much faster access to the
 #' full peak data (i.e. the numerical matrices of m/z and intensity values).
 #' Also subsetting and access to any spectra variable (except `"mz"` and
-#' `"intensity"` is fastest for the `MsBackendMemory`. Finally, the
-#' `MsBackendMemory` supports also arbitrary peak annotations while the
-#' `MsBackendDataFrame` does not have support for such additional peak
-#' variables.
+#' `"intensity"` is fastest for the `MsBackendMemory`.
```
lgatto, line 549: I think there's a missing ) after "intensity".

R/MsBackend.R
```diff
@@ -598,13 +590,12 @@
...
 #' Additional columns are allowed too.
 #'
-#' For the `MsBackendMemory`, any column in the provided `data.frame` which
-#' contains a `list` of vectors each with length equal to the number of peaks
-#' for a spectrum will be used as additional *peak variable* (see examples
-#' below for details).
+#' The `peaksData` function for `MsBackendMemory` and `MsBackendDataFrame`
+#' returns a `list` of `numeric` `matrix` by default (with parameter
+#' `columns = c("mz", "intensity")`). If other peak variables are requested,
+#' a `list` of `data.frame` is returned (to ensure m/z and intensity values
```
lgatto, line 596: Not sure to understand the "to ensure m/z ...". Should it not read "ensuring that m/z ..."?

R/MsBackendDataFrame.R
```diff
@@ -52,12 +55,16 @@ setMethod("show", "MsBackendDataFrame", function(object) {
...
 setMethod("backendInitialize", signature = "MsBackendDataFrame",
-          function(object, data, ...) {
+          function(object, data, peaksVariables = c("mz", "intensity"), ...) {
               if (missing(data)) data <- DataFrame()
               if (is.data.frame(data))
                   data <- DataFrame(data)
               if (!is(data, "DataFrame"))
                   stop("'data' has to be a 'DataFrame'")
+              peaksVariables <- intersect(peaksVariables, colnames(data))
+              if (sum(c("mz", "intensity") %in% peaksVariables) == 1L)
```
lgatto, line 65: If I understand this check, it will throw an error of only one of "mz" or "intensity" are in the peaksVariables. If none are, this wouldn't trigger the error. Why not 

```
sum(c("mz", "intensity") %in% peaksVariables) != 2L
```

R/MsBackendMemory.R
```diff
@@ -352,57 +357,51 @@ setReplaceMethod("mz", "MsBackendMemory", function(object, value) {
...
-                            b[, pdcol, drop = FALSE])[, columns])
-        }, SIMPLIFY = FALSE)
-    } else list()
-})
+setMethod(
+    "peaksData", "MsBackendMemory", function(object,
+                                             columns = c("mz", "intensity")) {
+        if (length(object)) {
+            cns <- colnames(object@peaksData[[1L]])
+            if (length(columns) == length(cns) && all(cns == columns))
```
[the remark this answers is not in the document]
lgatto: ok

R/MsBackend.R
jorainer: yes, thanks

R/MsBackendDataFrame.R
jorainer: good catch! thanks, changed that.

jorainer: actually, after checking again: we were supporting the following situations:

- user provides m/z and intensity
- user does not provide neither m/z and intensity (in which case 0-length spectra are available, but the `Spectra` is still considered valid).

An error is only thrown if either m/z but not intensity, or intensity but not m/z is provided. from a data point of view that does not make sense, you can't have one of the two. you could either have both or none.

Open to discuss @lgatto when you think that's not a good idea.

lgatto: OK, fine by me. May be just clarify with a comment?

jorainer: jepp, will add that and also check that I mention it in documentation

lgatto (approved):

R/MsBackendDataFrame.R
jorainer: added as a comment and docu.

Issue #293, opened 2023-05-25
Add a filterPrecursorPeaks function


jorainer:
Add a function `filterPrecursorPeaks` that removes peaks with m/z matching the precursor m/z from (fragment) spectra. An additional parameter `remove` could be used to specify which peaks should be removed: only matching peaks `"=="`, precursor peaks and all above `">="`, only those above `">"`, precursor peaks and all below `"<="`, only those below `"<"`. The function definition would be:

```r
filterPrecursorPeaks <- function(x, ppm = 20, tolerance = 0, remove = c("==", ">=", ">", "<=", "<"))
```

lgatto:
So this a convenience function that is similar to `filterMzValues()` but that operates on MS levels > 1 (i.e. those that have a precursor) and automatically considers uses the precursor m/z. This is indeed useful, but I am wondering about the many `remove` options other than `==`. What are the use cases? 

Coming back to the existing `filterMzValues()` and `filterMzRange()`. If I remember well, these are provided with fixed values or ranges, that are applied to all spectra of a given MS level. Wouldn't want a way to be able to pass different MZ values/ranges for different scans, which would then also address the need for a new `filterPrecursorPeaks()`?

jorainer:
Use cases for situations other than `"=="` (which removes the precursor peak itself from the fragment spectrum): `">="` for small molecule annotations this becomes more and more of an option (e.g. the spectral entropy similarity score does this by default): remove the precursor peak and all peaks with an m/z larger than the precursor from the spectrum (for single charged molecules it's not clear what these larger peaks actually are). The `">"` would then keep the precursor and simply remove all peaks larger than the precursor. This, along with `"<"` and `"<="` are just to provide all possibilities.

And yes, I will have a look into `filterMzRange` - maybe we can find a way to (re)use that function also for this use case.

jorainer:
I had a look into the `filterMzRange` and it is not (easily) possible to include also the precursor m/z of each spectrum into the m/z range to filter each spectrum individually.

Also, from the user perspective I believe it might be easier to understand (and find) to have a `filterPrecursorPeaks` function since the task is to remove potential precursor peaks from each spectrum (or precursor peaks and all peaks with an m/z >= the precursor m/z. So, I would go for the original option.

lgatto:
OK for the original option.

PR #299, opened 2023-06-28
Filter precursor peaks

jorainer:
Add `filterPrecursorPeaks` method to remove peaks within each spectrum based on their m/z values relative to the precursor m/z. See also issue #293 for a discussion.


sgibb (approved):
Thanks, everything fine!

Issue #303, opened 2023-11-10
`bin` function *create* peaks of intensity 0 


philouail:
The `bin` function when used in combination with `combineSpectra` will *create* peaks of intensity 0. This as for consequence to overload memory on big dataset, and instead of decreasing the complexity of spectra, it will increases it. 

Here is a reproducible example: 

```r
> data <- readMsExperiment(system.file("microtofq", "MM14.mzML", package = "msdata"))

> bps <- spectra(data[1]) |>
+     combineSpectra(intensityFun = max, ppm = 5)

> lengths(bps)
[1] 82696


> bps_bin <- spectra(data[1]) |>
+     bin(binSize = 0.01) |> 
+     combineSpectra(intensityFun = max, ppm = 10) 

> lengths(bps_bin)
[1] 90601
```
So 8,000 peaks were *created*

```r
> tbl <- bps |>
+     intensity()
> sum(tbl == 0)
C:\\Users\\plouail\\AppData\\Local\\R\\win-library\\4.3\\msdata\\microtofq\\MM14.mzML 
                                                                                    0 
> tbl_bin <- bps_bin |>
+     intensity()  
> sum(tbl_bin==0)
C:\\Users\\plouail\\AppData\\Local\\R\\win-library\\4.3\\msdata\\microtofq\\MM14.mzML 
                                                                                31307 
``` 

I think the function could be changed by *dropping* peaks of intensity 0.

jorainer:
Actually, I think the issue is the `bin` function itself:

```r
> lengths(spectra(data)) |> head()
[1] 1378 1356 1404 1496 1525 1498
> res <- bin(spectra(data), binSize = 0.01)
[1] 91100 91100 91100 91100 91100 91100
```

So, I would suggest the fix should go into `bin`. Maybe with an additional parameter `dropZero = FALSE` that, if set to `TRUE` drops all 0 intensity peaks. For the name of the parameter, I'm not quite sure if that `dropZero` is the best choice. Maybe `zero.rm` (similar to `na.rm`)?

jorainer:
My reasoning why I think adding a parameter instead of just removing zero-intensity peaks by default: the current default would allow to directly `rbind` the intensity values returned by `bin,Spectra` which again would allow *simple* visualizations of multiple spectra at once or to use e.g. standard similarity scoring on such matrices (see [here](https://rformassspectrometry.github.io/Spectra/articles/Spectra.html#comparing-spectra)) for an example).

[...]

PR #305, opened 2023-11-20
fix: `bin` function

philouail:
This is related to issue #303, this pull request add the argument `zero.rm` with `TRUE` as a default. 

The code and documentation updated. 
I had to fiddle with the test files because the tests were made with the assumptions that the 0 would be kept. 
For now I just added zero.rm = FALSE and added code chunk to check that the bins of intensity 0 are being removed. 
See the comment for one thing that I am a bit confused about. 

I can also change all the tests if you think it's better.


philouail:
The vignette is updated and the unit test are all good for the `bin` functions. There is some failure with the plotting functions but I guess that's something from before ? 

So the PR is ready for merging !

jorainer (approved):
Thanks Phili! Great PR.

Issue #304, opened 2023-11-15
Memory issues and (potential) chunk-wise processing of peaks data


jorainer:
By using a `MsBackendSql` for a very large MS experiment we ran into memory issues (see issue #303 ) processing the data. All operations working on peaks data use the internal `.peaksapply` function that by default splits the processing based on `dataStorage(x)` and performs the operation in parallel (even if *processing* might only mean loading the data from the raw data files). For in memory backends and the e.g. `MsBackendSql`, `dataStorage` will only have a single value, thus no splitting/parallel processing is performed.

Advantages of splitting and parallel processing

- performance: for computationally intense processing steps.
- lower memory demand: only the data currently being processed would be loaded into memory.

These advantages are good for very large data sets. The downside of splitting and parallel processing is however also the overhead of splitting and combining the data. Thus, in many cases (e.g. also for in-memory backends) it might actually be good to **not** do it.

Maybe it would make sense to add a parameter or let the user choose to enable parallel/split processing, even if backends other than `MsBackendMzR` are used.

jorainer:
A possibility would be to add a specific parameter for this to any of the user-facing functions: `mz`, `intensity`, `peaksData`, `bin`, `ionCount`, `lengths` etc.

A parameter `chunkSize = integer()` could be used. If `length(chunkSize)`, a factor could be created to be passed to parameter `f` of the internal `.peaksapply`, hence forcing to perform splitted (eventually parallel) processing. With the default `chunkSize = integer()` per-file parallel processing would be performed for `Spectra` with different levels for `dataStorage` (as it is currently the default).

jorainer:
An example for the high memory requirement: determine the number of peaks per spectrum from a `Spectra` using a `MsBackendSql` backend - limiting to 500,000 spectra to avoid running out of memory:

```r
> peakRAM(tmp <- lengths(sps[1:500000]))
               Function_Call Elapsed_Time_sec Total_RAM_Used_MiB Peak_RAM_Used_MiB
1 tmp<-lengths(sps[1:5e+05])           63.094                6.4           32650.3
```

Thus, to complete this simple operation 32GB of memory are required, because the full peaks data is first loaded into memory before the number of peaks is determined.

jorainer:
After some tests and evaluating the code: passing parameters to individual functions might not be the ideal solution (also because some functions don't allow that, such as `lengths`).

[...]

PR #306, opened 2023-11-22
Enable chunk-wise processing for all peaks data functions

jorainer:
This PR fixes issue #304 . In brief: it adds the possibility for the user to set and define chunk-wise processing of a `Spectra`. This will affect all functions working on peaks data (e.g. even `lengths`, `mz`, `peaksData`) and ensures that even large-scale data can be handled reducing out of memory errors.

What this PR adds:

- `Spectra` gains a new slot `@processingChunkSize`.
- Add functions `processingChunkSize` and `processingChunkSize<-` to get or set the size of chunks for chunk-wise processing. The default is `Inf` hence no chunk-wise processing is performed (important e.g. for small data sets of in-memory backends).
- Add `backendParallelFactor,MsBackend` method: this allows backends to *suggest* a preferred splitting of the data into chunks. The default is to return `factor()` (i.e. no preferred splitting), `MsBackendMzR` on the other hand returns a `factor` depending on the `"dataStorage"` spectra variable (hence suggests splitting by original data file).
- The internal `peaksapply` function uses either chunks defined through `processingChunkSize` for chunk-wise processing, or if not set, uses the suggested splitting from the backend (through `backendParallelFactor`).
- The user can check if and how the `Spectra` will be split using the `processingChunkFactor` function that returns a `factor` representing the chunks (defined through `processingChunkSize`), or, if not set, the suggested splitting (through `backendParallelFactor`) or `factor()` in which case no chunk-wise processing is performed.

This processing is used for all `Spectra` methods accessing peaks data (or processing the peaks data). To avoid performance loss for small data sets or in-memory backends it is not performed by default. If enabled by the user, it allows to process also large data.

I think this is a very important improvement allowing the analysis of large (on-disk) data - for which we ran into unexpected issues (see #304). 

Happy to discuss @sgibb @lgatto @philouail .


vignettes/MsBackend.Rmd
```diff
@@ -96,12 +96,21 @@ General conventions for MS data of a `Spectra` are:
...
+parameter `processingChunkSize`. Alternatively, through the
+`backendParallelFactor` function the backend can also *suggest* a `factor` that
+should/could be used for splitting and parallel processing. The default
+implementation for `backendParallelFactor` is to return an empty `factor`
+(`factor()`) hence not suggesting any preferred
+splitting. `backendParallelFactor` for `MsBackendMzR` on the other hand returns
+a `factor` based on the data files the data is stored in (i.e. based on the
+`dataStorage` of the MS data).
+
+Besides parallel processing, this chunk-wise processing can also reduce the
```
philouail, line 111: I know it's very much implied but here or in the chunk processing vignette maybe explicitly say that it is advised to set up for big dataset that do not have `MsBackendMzR` type of backend. Just so that less experienced user get the important info.

R/MsBackend.R
```diff
@@ -849,6 +861,13 @@ setMethod("backendMerge", "MsBackend", function(object, ...) {
     stop("Not implemented for ", class(object), ".")
 })
 
+#' @exportMethod backendParallelFactor
+#'
+#' @rdname MsBackend
+setMethod("backendParallelFactor", "MsBackend", function(object, ...) {
+    factor()
```
philouail, line 868: Trying to visualize what happen between different uses, but essentially why do we need this method ? Because it will only be used by users that have datasets with `MsBackendHdf5Peaks` or `MsBackendMzR` no ? Or is it just mandatory to have a default implementation ? you cannot just leave the function throwing an error if user try to use this function with other backends ?

philouail: actually yes, why isn't it define exactly like the method below for `export` ?

R/Spectra-functions.R
```diff
@@ -1008,3 +1046,138 @@ filterPrecursorPeaks <- function(object, tolerance = 0, ppm = 20,
...
+
+#' @title Define a factor for parallel access to peaks data
+#'
+#' @description
+#'
+#' Define a factor to split a `Spectra` for parallel processing based on
+#' parallel processing setup,
+#'
+#' If `processingChunkSize(x)` is defined and its value is smaller then
+#' `length(x)`, a factor depending on this chunk size is returned.
```
philouail, line 1058: Why `length(x)` ? and not `max(lenghths(x))` ? So that if the user wants to use a chunksize smaller than the number spectra per file they can but otherwise do it files by files.. idk if that make sense... 
And i realise that only really make sense if the backend is `MsBackendMzR`...

philouail: I know my comment does not make much sense, i probably misunderstood something here aha, and I was just a bit confused by the `lenght(x)`

R/Spectra.R
```diff
@@ -1723,50 +1690,43 @@ setMethod("dropNaSpectraVariables", "Spectra", function(object) {
     object
 })
 
+test_fun <- function(x) {
```
philouail, line 1693: What is this ?

R/MsBackend.R
jorainer: the idea is that with this method we allow the backend to suggest an (optimal) factor to split the data. Here we define a default implementation for **any** `MsBackend` class, that simply returns `factor()`, i.e. does not suggest any splitting.  For `MsBackendMzR` and `MsBackendHdf5Peaks` we overwrite this method to actually suggest a splitting by original data file (or rather data storage). 

We will call `backendParallelFactor(object@backend)` (`object` being a `Spectra`) for some data access or processing methods to evaluate whether we need to split the data for chunk-wise processing. If we would throw an error like below every `MsBackend` implementation would need to implement this method (even if it will return `factor()` like here with the default implementation). Having an actual default implementation here reduces the code that needs to be implemented for each backend (and also reduces code repetition, because all backends for now, except the two mentioned above) will not provide any suggested splitting.

R/Spectra-functions.R
jorainer: The difference between `length` and `lenghts` is a bit tricky - so, `length(x)` is the number of spectra (returns a single value) while `lengths(x)` is the number of mass peaks per spectrum (returns a vector of length equal to the number of spectra). We want to split and chunk-wise process the `Spectra` object, it makes thus only sense to split if the chunk size is smaller then the length (number of spectra) or a `Spectra`. This is independent of the number of mass peaks per spectrum, we just want to process a smaller subset of our `Spectra` at a time and load only the data for that to reduce the memory demand.

vignettes/MsBackend.Rmd
jorainer: yes! thanks! I will also add a dedicated vignette on that topic.

jorainer:
@philouail , I've fixed some more things, can you please give again a careful look - any questions, concerns, comments or change requests highly welcome!

I've added now also a vignette describing the parallel processing settings. Please have a look at that too.

philouail (approved):
This looks super good to me. 
The `processingChunkSize` and `processingChunkFactor` description is really really good. 
The vignette with the tips for large dataset is super good and straight to the point i like it. 

I made very few comments and I had one thing I was confused about.

R/Spectra-functions.R
```diff
@@ -1008,3 +1046,138 @@ filterPrecursorPeaks <- function(object, tolerance = 0, ppm = 20,
...
+#'
+#' Some backends might not support parallel processing at all.
+#' For these, the `backendBpparam` function will always return a
+#' `SerialParam()` independently on how parallel processing was defined.
+#'
+#' @param x `Spectra`.
+#'
+#' @param value `integer(1)` defining the chunk size.
+#'
+#' @return `processingChunkSize` returns the currenlty defined processing
```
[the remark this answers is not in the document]
jorainer: thanks! I fixed it.

R/Spectra.R
```diff
@@ -1657,25 +1631,19 @@ setMethod("acquisitionNum", "Spectra", function(object)
...
-        BPPARAM <- backendBpparam(object, BPPARAM)
-        if (is(BPPARAM, "SerialParam"))
-            f <- NULL
-        else f <- factor(dataStorage(object))
-        SimpleList(.peaksapply(object, f = f, columns = columns,
-                               BPPARAM = BPPARAM, ...))
-})
+        f <- .parallel_processing_factor(object)
+        if (length(object@processingQueue) || length(f))
+            SimpleList(.peaksapply(object, columns = columns, f = f))
```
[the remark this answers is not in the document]
jorainer: If we don't specify `f` `.peaksapply` will use the default. For some methods such as `intensity` we could directly call the implementation from the backend bypassing `.peaksapply` - for that I however need to first check whether the processing queue is empty, or that chunk-wise processing is disabled. For that I call e.g. `f <- .parallel_processing_factor(object)` and, if we perform chunk-wise processing, we pass this parameter directly to the `.peaksapply` function (thus we can avoid that `f` needs to be again defined by another `f <- .parallel_processing_factor(object)` - which is the default for this parameter in `.peaksapply`). Hope this clarifies it a bit?

R/Spectra-functions.R
```diff
@@ -91,11 +91,40 @@ NULL
...
+#' - `$`.
+#' - `bin`.
+#'
+#' In *Spectra-functions.R*:
+#'
+#' - `.compare_spectra_chunk`: has its own `chunkSize` parameter and ignores
+#'   any other splitting. Does also some weird internal chunking by
+#'   `dataStorage` that we might want to skip/disable (or use
+#'   `backendParallelFactor` instead).
+#' - `.compare_spectra_self`: does not do any splitting.
```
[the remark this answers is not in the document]
jorainer: these should be OK-ish. but there are the `containsMz` and the `containsNeutralLoss` methods in *Spectra.R* that should get an update to use the same concept like e.g. `ionCount` or similar. I could open an issue and assign it to you :)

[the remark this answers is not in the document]
philouail: Perfect I'll check that next week :)

R/Spectra.R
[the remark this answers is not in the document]
philouail: Yes ok I think I understand thanks !

andreavicini (approved):
Seems very good to me!

vignettes/Spectra.Rmd
```diff
@@ -41,7 +41,9 @@ This vignette provides general examples and descriptions for the *Spectra*
 package. Additional information and tutorials are available, such as
 [SpectraTutorials](https://jorainer.github.io/SpectraTutorials/),
 [MetaboAnnotationTutorials](https://jorainer.github.io/MetaboAnnotationTutorials),
-or also in [@rainer_modular_2022].
+or also in [@rainer_modular_2022]. For information how to handle and (parallel)
```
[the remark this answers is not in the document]
jorainer: thanks!

jorainer:
Thanks for the reviews @andreavicini and @philouail ! I will merge now after having another look myself again.

sgibb:
@jorainer sorry but I didn't review the code but a small suggestion anyway: If we add a new slot to the `Spectra` class it would break backward compatibility. So I would suggest to increment the "version" of the class and increment the minor number of the version field in the DESCRIPTION file.

jorainer:
Hi Sebastian @sgibb , thanks for the suggestion. I'll increment the class version. I ensured backward compatibility through the accessor function that checks if the object has the slot and if not returns `Inf` (the default for the slot value). Also, `Spectra` methods will (automatically) call `updateObject` if required. So, backward compatibility should be guaranteed.

I would maybe not bump the minor version of the package to not interfere with the Bioconductor versioning?

PR #309, opened 2024-01-23
add entropy functions

mar-garcia:
@jorainer, I've tried to follow your suggestions... 
Not sure if I've understood everything 100%, but let's try to star with what I've done here :) 
Let me know if I should proceed in another way, change anything, etc..
Thanks!!


R/Spectra.R
```diff
@@ -2639,3 +2642,15 @@ setMethod("combinePeaks", "Spectra", function(object, tolerance = 0, ppm = 20,
         ppm, " and tolerance = ", tolerance, ".")
     object
 })
+
+
+#' @rdname Spectra
+#'
+#' @importFrom MsCoreUtils entropy
+setMethod("entropy", "Spectra", function(object, normalized = TRUE) {
+  if (length(object))
```
jorainer, line 2651: for `normalize == TRUE` we would need to call `nentropy`, for `normalized = FALSE` `entropy` - I would suggest to add something like:

```r
if (normalized) entropy_fun <- nentropy
else entropy_fun <- entropy
unlist(.peaksapply(
    object, FUN = function(pks, ...) entropy_fun(pks[, "intensity"])), 
    use.names = FALSE
))
```

ah, and please remove the `na.rm = TRUE` since the `entropy` function does not support that.

R/Spectra.R
```diff
@@ -2639,3 +2642,15 @@ setMethod("combinePeaks", "Spectra", function(object, tolerance = 0, ppm = 20,
         ppm, " and tolerance = ", tolerance, ".")
     object
 })
+
+
+#' @rdname Spectra
+#'
+#' @importFrom MsCoreUtils entropy
```
jorainer, line 2649: here import both, entropy and nentropy

jorainer (changes requested):
Thanks Mar for the PR!

I have some change requests and I would also ask you to add yourself as a contributor to the DESCRIPTION (similar to https://github.com/rformassspectrometry/MsCoreUtils/blob/main/DESCRIPTION), to also increase the version in DESCRIPTION to 1.13.4 and to also add an entry to the NEWS file mentioning that you added the `entropy` method.

jorainer (approved):

jorainer:
Thanks Mar @mar-garcia for this contribution! I'll merge and fix the missing NAMESPACE entry.

PR #314, opened 2024-03-01
updated filepath in filterFourierTransformArtefacts example

stanstrup:
the filepath in the example is no longer valid


jorainer (approved):
Thanks Jan!

Issue #315, opened 2024-03-06
Adjust precursor m/z values for MS2 spectra from Waters instruments


jorainer:
Originally mentioned by @mar-garcia (also please udate/fix if I don't describe the problem properly).

Waters instruments inject the calibrant every x scans. Independently of the whether the data is calibrated online (during acquisition) or offline (afterwards by Waters software or Proteowizard), m/z values within spectra are correctly calibrated, but the precursor m/z of MS2 spectra **is not**.

The suggestion from Mar how to correct this (for DDA data):

- for each fragment spectrum, get the most recent previous MS1 spectrum.
- select mass peaks within a certain m/z tolerance to the reported precursor m/z in that MS1 spectrum
- extract the m/z of the mass peak with the highest intensity from the pre-selected MS1 mass peak and use that as the *adjusted* precursor m/z.

jorainer:
Maybe as a cross-check (or if no MS1 data is available), could it make sense to also check the actual fragment spectrum and check if there are fragment peaks close to the reported precursor m/z value and use one with a high intensity? - assuming that we also can detect the precursor as a fragment...

mar-garcia:
Yes, it can be an option in case no MS1 data is available, but in some cases (not sure if in most of cases, but I would say that in a relevant proportion of cases...) the "risk" of not having the precursor m/z in the fragment spectrum is quiet high I would say.... (I mean, the intstrument is "completly" fragmenting the precursor), also depening on the collision energy applied I would say....

jorainer:
Hey @mar-garcia , can you please test if this function works for you? You can install the package with `BiocManager::install("RforMassSpectrometry/Spectra", ref = "adjust_precursor_mz")` - eventually you might also need to update `ProtGenerics` using `BiocManager::install("RforMassSpectrometry/ProtGenerics")`.

Assuming `s` is a `Spectra` you can adjust/estimate the precursor m/z using `pmz <- estimatePrecursorMz(s)` - adapting eventually also parameters `tolerance` and `ppm`.

mar-garcia:
Hi!
Just checked and I think it's working!
Thanks a lot!!! :)

jorainer:
Closing as functionality was now added.

PR #317, opened 2024-03-20
feat: add estimatePrecursorMz function (issue #315)

jorainer:
- Add the `estimatePrecursorMz` functions to estimate/adjust the reported precursor m/z for DDA spectra.

This PR fixes issue #315 , i.e. for DDA data from Waters instruments, the reported precursor m/z value might not be correct because it is not being corrected through the lockmass calibration. Would also be good to get hold of some more (different) Waters DDA files to check.


R/Spectra-estimatePrecursorMz.R
```diff
@@ -0,0 +1,137 @@
...
+#' spectrum, this method identifies the peak with the highest intensity and
+#' an m/z value similar to the fragment spectrum's reported precursor m/z
+#' (given parameters `tolerance` and `ppm`). This m/z value is then reported.
+#' Since the fragment spectrum's potential MS1 mass peak is selected based on
+#' its intensity, this method should **only be used for DDA data**.
+#'
+#' @note
+#'
+#' This approach should only be applied if fragment spectra are recorded
+#' through  data dependent acquisition (DDA), since only with this data it
```
philouail, line 26: there's an extra space between "through" and "data".

philouail: Also I'm a bit confused by the sentence, maybe rephrasing would help: 

"This approach is applicable only when fragment spectra are obtained through data-dependent acquisition (DDA), as it assumes that the peak with the highest intensity within the given m/z window corresponds to the precursor ion."

philouail (approved):
All good to me, super cool function and good that this issue was discovered !  I was just confused by one paragraph in the documentation, I put an alternative in case.

R/Spectra-estimatePrecursorMz.R
jorainer: Oh yes, great, thanks! that's much better. I'll fix.

jorainer:
@mar-garcia , can you please check the documentation of the function? just to ensure it's correct and describes the problem properly.

mar-garcia:
Not sure how I can check the documentation.... I've just checked what is written here: https://github.com/rformassspectrometry/Spectra/blob/adjust_precursor_mz/R/Spectra-estimatePrecursorMz.R
The only small detail I would add the specifcation "in the previous MS1 spectrum" in the following sentence:
Thus, for each fragment spectrum, this method identifies **_in the previous MS1 spectrum_** the peak with the highest intensity and an m/z value similar to the fragment spectrum's reported precursor m/z (given parameters `tolerance` and `ppm`).

jorainer:
that's perfect, thanks @mar-garcia !

Issue #321, opened 2024-05-15
Add possibility to change the file path for a `MsBackendMzR`


jorainer:
We have some use cases in which we run some part of the analysis (e.g. the preprocessing using `xcms`) on our cluster, save then the results (which include a `Spectra`) to an RData to then continue the analysis (more interactively) on a local computer. This works only if we use a `MsBackendMemory` as a backend, but e.g. with a `MsBackendMzR` we would have different file paths locally and on the cluster.

Maybe we could add a method `backendFilePath,MsBackend` and `backendFilePath<-,MsBackend` that allows to get/change the directory where the data is stored - for backends that support this.

A method could be implemented for `MsBackendMzR`, `MsBackendHdf5Peaks` and eventually the `MsBackendOfflineSql`, if a SQLite database is used. Other backends would throw an error as this method would not be supported/needed for them.

Open for discussions or alternative ideas/thoughts.

jorainer:
Pinging @philouail

jorainer:
Maybe `dataStorageBasePath` would make more sense as a name - so, essentially, the common (base) path for the `dataStorage`. For *in-memory* backends or backends where the `dataStorage` is not defined or used it could/should return `NA_character_`, while for backends that keep a reference to the (on-disk stored) data it should return the base file path common to all spectra in `Spectra` (as determined with `MsCoreUtils::common_path()`).

PR #323, opened 2024-05-17
Add a method to change the data storage base path for on-disk backends

jorainer:
On-disk backends depend on the (original) data files being available. Moving/copying serialized `Spectra` objects with such backends between file systems or moving the original data files will break their functionality.

This PR adds a new method to the `MsBackend`, `dataStorageBasePath()` and `dataStorageBasePath<-` to get or set the base path to the directory of the data storage files for on-disk backends (issue #321 ).


philouail (approved):
It looks all good to me and I did not get problems with windows system. 

Thanks super helpful !!

PR #330, opened 2024-09-13
Improve and fix the Spectra constructor method

jorainer:
Update the `Spectra()` constructor method to better support initialization with backends defining their own specific parameters. This also fixes https://github.com/rformassspectrometry/MsBackendMetaboLights/issues/4


jorainer:
Note: I'll directly merge without review.

PR #331, opened 2024-09-25
Restructure documentation

jorainer:
- Organize `Spectra` documentation into thematically connected documentations.
- Add aliases for all methods to avoid e.g. `?intensity` ending up in the documentation in *ProtGenerics*.


jorainer:
This fixes #288

jorainer:
@philouail , no need to go through the full documentation - maybe just some feedback if you're OK with the structure now or if it would make sense to further divide the documentation into smaller parts.

philouail (approved):
Not the easiest to read in Github I think aha. But this looks great, i'm super happy, thanks !

PR #332, opened 2024-09-27
Add new extractByIndex method

jorainer:
This PR adds a new generic and method `extractByIndex()` to subset/extract data from a `MsBackend` given `integer` indices. Properties:

- `[` method calls `extractByIndex()`, converting first parameter `i` to an `integer`.
- analysis methods for `Spectra` use now `extractByIndex()` to subset the backend instead of `[`:
  - this fixes issue https://github.com/rformassspectrometry/MsBackendMetaboLights/issues/5
  - there is a small performance gain to, because of less input argument checking.

Disadvantage:
- each `MsBackend` implementation **must** implement this method now. So, I'll need to fix the other backend classes/packages accordingly.

Open to discussion.


jorainer:
Update: just added a backward compatibility that fails back to use `[` if `extractByIndex()` is not implemented for a backend.

vignettes/MsBackend.Rmd
```diff
@@ -563,35 +563,39 @@ additionally available variables and the `columns` parameter of the
...
+index, while `[`, to be compliant with the base R implementation, should support
+to subset by indices or logical vectors. An error should be thrown if indices
+are out of bounds, but the method should also support returning an empty backend
+with `[integer()]`. Note that the `MsCoreUtils::i2index` function can be used to
 check for correct input (and convert the input to an `integer` index).
 
-Below we implement a possible `[` for our test backend class. We ignore the
-parameters `j` from the definition of the `[` generic, since we treat our data
-to be one-dimensional (with each spectrum being one element).
+The `extractByIndex()` method is used by the data operation and analysis methods
```
philouail, line 579: just so I understand, from now on when we need to implement some code in Spectra we have to use `extractByIndex()` instead of `[` ?

philouail (approved):
Looks good to me, a bit annoying you had to get to such a big change to fix something like that ! Thanks for figuring it out ! I have one question in comment. 

Did you get any answer from the S4Vectors issue you made ?

vignettes/MsBackend.Rmd
jorainer: we can use both - but it would be *better* to use `extractByIndex`.

Issue #342, opened 2024-11-21
Add a function to append multiple spectra variables to a `Spectra`


jorainer:
We have already the `joinSpectraData()` function, but I would still like to have a different (maybe easier?) way of adding new spectra variable *columns* to a `Spectra` object. With `[[<-` and `$<-` it's already possible to add a single new spectra variable (or also to replace an existing variable), but we have frequently a `data.frame` or `DataFrame` (or `matrix`) with multiple columns I would like to add.

I would thus propose to have an additional function, e.g. `cbind,Spectra,ANY <- function(x, y)` that allows to add data for multiple values in one go. The properties would be:

- `nrow(y)` has to be same as `length(x)`
- throw an error if `x` has already a spectra variables with a name, to avoid replacing values, but just adding new values.

ideally, this should be added as a new method for `MsBackend` with a default implementation that (re)uses the `[[<-` method and a loop - some backends, such as `MsBackendMemory` could implement their own version that directly calls `cbind` on its `@spectraData` `data.frame`.

@lgatto @sgibb , any thoughts?

ping @philouail

lgatto:
That sounds like a useful thing to have, indeed. 

`nrow(y)` has to be same as `length(x)` is equivalent to `nrow(y) is identical to `nrow(spectraData(x))`. Any consideration regarding regarding `rownames(x)`, or some mechanism to insure that the rows or `y` and `spectraData(x)` are matching when `cind`ing?

jorainer:
yes, `nrow(y)` is equivalent to `nrow(spectraData(x))` - but a backend can have a more efficient `lenght()` method implemented that does not require to first get the spectra data with the `spectraData(x)` call.

I would actually keep the function simple and don't do any checks on rownames - or change row names. we don't explicitly support or force row names on `spectraData()`, as e.g. if you have a SQL backend, there are no row names available.

So, `cbind()` should *blindly* append *columns* (i.e. `spectraVariables`) to a `Spectra` assuming the user provides the rows in the correct order (similar to a `$<-` call). I would throw an error if a) `nrow()` does not match the number of spectra in `Spectra` or if any of the colnames in `y` are already `spectraVariables()`.

And actually, I would maybe implement `cbind2()` instead of `cbind()` - then we have a bit better control over the dispatching.

PR #343, opened 2024-12-04
Addition of cbind2() to append multiple spectra variables to the spectra data

philouail:
Imported `cbind2(x,y,...)` method from `methods` and implemented a default method for `MsBackend`. As well as a more efficient one for `MsBackendMemory` and `MsBackendDataFrame`. 

See #342 for discussion about this.


R/MsBackend.R
```diff
@@ -182,10 +182,14 @@
 #' @param value replacement value for `<-` methods. See individual
 #'     method description or expected data type.
 #'
-#' @param values for `filterValues()`: A `numeric` vector that define the
+#' @param values For `filterValues()`: A `numeric` vector that define the
 #'     values to filter the `object`. `values` needs to be of same length than
 #'     parameter `spectraVariables` and in the same order.
 #'
+#' @param y For `cbind2()`: A `data.frame` or `DataFrame` with the
+#'     spectra variables to be added to the backend. Need to be of the same
```
jorainer, line 190: I would maybe be a bit more precise here:

*The number of rows and their order has to match the number of spectra in `x`, respectively their order.* instead of *Need to be ...*. Or something along these lines, but mention that a) the number of rows has to match AND that the order is important.

R/MsBackend.R
```diff
@@ -313,6 +317,11 @@
 #'   `dropNaSpectraVariables()` might still show columns containing `NA` values
 #'   for *core* spectra variables.
 #'
+#' - `cbind2()`: allows to appends multiple spectra variables to the backend at
+#'   once. It does so *blindly* and is therefore at the risk of the user. For a
```
jorainer, line 321: Maybe describe what you mean with *blindly* as that might be confusing to the user. Maybe rather say that *the `Spectra` and the values for the new spectra variables have to be in a matching order*. Do we actually allow to replace also spectra variables? if not we should also mention that here.

R/MsBackend.R
```diff
@@ -1022,6 +1031,26 @@ setMethod("peaksVariables", "MsBackend", function(object) {
...
+#'
+#' @importMethodsFrom methods cbind2
+#'
+#' @rdname MsBackend
+setMethod("cbind2", signature = c("MsBackend", "dataframeOrDataFrameOrmatrix"),
+          function(x, y = data.frame(), ...) {
+    if (is(y, "matrix"))
+        y <- as.data.frame(y)
+    if (nrow(y) != length(x))
+        stop("Length of 'y' does not match the number of spectra in 'x'")
```
jorainer, line 1046: better say here *'nrow(y)' does not...* - because the `length()` of a `data.frame` is actually the number of columns, so it could be a confusing error message.

R/MsBackend.R
```diff
@@ -1022,6 +1031,26 @@ setMethod("peaksVariables", "MsBackend", function(object) {
...
+#' @exportMethod cbind2
+#'
+#' @importMethodsFrom methods cbind2
+#'
+#' @rdname MsBackend
+setMethod("cbind2", signature = c("MsBackend", "dataframeOrDataFrameOrmatrix"),
+          function(x, y = data.frame(), ...) {
+    if (is(y, "matrix"))
+        y <- as.data.frame(y)
+    if (nrow(y) != length(x))
```
jorainer, line 1045: should we also check if spectra variables already exist? That's open for discussion, but maybe it would be safer if `cbind2` would allow to only append new columns instead of replacing existing columns?

R/MsBackendDataFrame.R
```diff
@@ -567,6 +567,23 @@ setMethod("[", "MsBackendDataFrame", function(x, i, j, ..., drop = FALSE) {
...
+#' @importMethodsFrom methods cbind2
+#'
+#' @rdname hidden_aliases
+setMethod("cbind2", signature = c("MsBackendDataFrame",
+                                  "dataframeOrDataFrameOrmatrix"),
+          function(x, y = data.frame(), ...) {
+              if (is(y, "matrix"))
+                  y <- as.data.frame(y)
+              if (nrow(y) != length(x))
+                  stop("Length of 'y' does not match the number of spectra in 'x'")
```
jorainer, line 581: Also here, please use *nrow* (or spell our *Number of rows*) instead of *Length*

R/Spectra.R
```diff
@@ -1669,6 +1687,16 @@ setMethod("c", "Spectra", function(x, ...) {
     .concatenate_spectra(unname(list(unname(x), ...)))
 })
 
+setClassUnion("dataframeOrDataFrame", c("data.frame", "DataFrame"))
```
jorainer, line 1690: why are you defining here this class union? in `MsBackend` you have already `data.frame`, `DataFrame` and `matrix` - I would re-use that actually.

jorainer (changes requested):
Really helpful addition @philouail ! 

I have some minor suggestions for the documentation and one point to discuss.

R/MsBackend.R
philouail: is the definition fine for you in Spectra.R ? Or should oi precise it  the same way ?

jorainer (approved):
Looks good! Thanks Phili!

PR #345, opened 2025-01-15
Changes of functions to method and move to ProtGenerics 

philouail:
The functions changes are the following: 

- applyProcessing()
- processingChunkFactor()
- processingChunkSize() and processingChunkSize<-

It was done so they could be implemented in the Chromatograms package.


jorainer (approved):
Looks good to me, thanks Phili!

Issue #346, opened 2025-01-21
Improve Spectra::plotSpectra and PSMatch::addFragments


guideflandre:
I am currently working on PTMs and variable modifications are a key aspect I wish to include in my workflow. 
As of the latest PR in PSMatch, `calculateFragments` now also measures a sequence's mass with variable modifications (see this [PR](https://github.com/rformassspectrometry/PSMatch/pull/16)). 

However, currently, `addFragments` and `plotSpectra` as well as  `plotSpectraOverlay` and `plotSpectraMirror`  do not support such modifications. There is also an error to be fixed with `addFragments` not allowing custom parameters when multiple spectra are called (see issue [here](https://github.com/rformassspectrometry/PSMatch/issues/18)).

Taking into account the already mentioned potential for optimised plots (see this [issue]), I revisited `plotSpectra` and `addFragments` to include:
- A [Universal Spectrum Identifier](https://www.nature.com/articles/s41592-021-01184-6)
- Fragmented sequence (with dividers) similar to the plots in [this paper](https://analyticalsciencejournals.onlinelibrary.wiley.com/doi/10.1002/pmic.202000061)
- Enable variable modifications
- Different ion types
- Color coded annotations
- Allow multiple plots with/without modifications (fix `addFragments` issue mentioned above)
- fix issues with `par()` going off when multiple plots are called
- Remove the redundant title (now in the USI) and reduce the unused space

The results look like this:
- [ ] A simple annotated plot: plotSpectra2(bulk[504], labels = addFragments2(bulk[504], type = c("a", "b", "y")), main = "A first example")

![Image](https://github.com/user-attachments/assets/e851c9f8-d88a-4774-821f-5018749c89b7)

- [ ] One single spectrum with variable moidifications: `plotSpectra2(sc[532], labels = addFragments2(sc[532], variable_modifications = c(D = 5)))`

![Image](https://github.com/user-attachments/assets/d9837559-4ca6-4053-a7c4-899965e421b7)

- [ ] Multiple plots with specific variable modification`plotSpectra2(sc[532:534], labels = addFragments2(sc[532:534], variable_modifications = c(D = 5)))`

![Image](https://github.com/user-attachments/assets/3a12b78e-7604-4d73-bb7d-1d02cdad531b)

- [ ] Fixed modifications can still be used (in the same way `modifications` parameter was used before. 

[...]

PR #348, opened 2025-01-22
Update plotSpectra environment to allow addFragments returned list

guideflandre:
This PR is linked to [this PSMatch issue](https://github.com/rformassspectrometry/PSMatch/issues/18) and the discussed topics in #346 

_The motivation for this PR:_
1) `addFragments` used to throw an error when multiple spectra where called
2) The new `variable_modifications` parameter necessitated a new `addFragments` function to distinguish annotations

_The solution:_
`addFragments` now returns a list of named elements (peptide sequence that includes modifications) containing a `character()` vector with the annotations. Each element of the list has an attribute linking the annotations to the spectrum it belonged to. In Spectra, the changes are located at the level of the presence of `labels`. It is assumed that most often than not, `addFragments` is called for `labels`.  A quick check for variable modifications is also added (as plotSpectra doesn't support them yet). All three functions (`plotSpectra`, `plotSpectraMirror` and `plotSpectraOverlay` now support the version of `addFragments` in [this PR](https://github.com/rformassspectrometry/PSMatch/pull/20)).

**This PR can only be accepted if [the corresponding PR in PSMatch](https://github.com/rformassspectrometry/PSMatch/pull/20) is accepted !!** Otherwise, `plotSpectra` won't work.


R/plotting-functions.R
```diff
@@ -217,10 +217,19 @@ plotSpectra <- function(x, xlab = "m/z", ylab = "intensity", type = "h",
...
+        mod_check <- sapply(labels, function(labels) attr(labels, "spectrumNumber"))
+        if (length(mod_check) != length(x))
+            stop("Variable modifications are not yet supported.")
+    } else {labels <- NULL}
+    
     for (i in seq_len(nsp))
         .plot_single_spectrum(x[i], xlab = xlab, ylab = ylab, type = type,
                               xlim = xlim, ylim = ylim, main = main[i],
-                              col = col[[i]], labels = labels,
+                              col = col[[i]], labels = labels[[i]],
```
jorainer, line 232: if I understand it, `labels` can be either 1 or equal to the length of `x` - so, I guess above you would need to ensure that if `labels` is of length 1 it gets repeated `length(x)` times - otherwise you would get here an out-of-bound error.

R/plotting-functions.R
```diff
@@ -217,10 +217,19 @@ plotSpectra <- function(x, xlab = "m/z", ylab = "intensity", type = "h",
         main <- rep(main[1], nsp)
     if (nsp > 1)
         par(mfrow = n2mfrow(nsp, asp = asp))
+    if (length(labels)) {
+        if (is.function(labels)) {
+            labels <- labels(x)
+        }
+        mod_check <- sapply(labels, function(labels) attr(labels, "spectrumNumber"))
```
jorainer, line 224: what is this line? why does `labels` need attributes?

R/plotting-functions.R
```diff
@@ -217,10 +217,19 @@ plotSpectra <- function(x, xlab = "m/z", ylab = "intensity", type = "h",
         main <- rep(main[1], nsp)
     if (nsp > 1)
         par(mfrow = n2mfrow(nsp, asp = asp))
+    if (length(labels)) {
+        if (is.function(labels)) {
+            labels <- labels(x)
```
jorainer, line 222: Why do you need to call a `labels()` function on the full `Spectra` object `x`? with the original code, the function would have been called on each spectrum individually (in the `.plot_single_spectrum()` function), so this would also avoid the need to support `labels` being a `list`.

In essence, the original `plotSpectra()` would call the `labels()` function on each individual spectrum, thus IMHO it would only require a `labels()` function that is able to extract whatever should be added/labeled from the spectra or peaks variables of that one spectrum. Would that not be possible in your case?

jorainer (changes requested):
Thanks for the PR! I have however some questions.

jorainer:
General comment: I don't know your specific use case, but would it not be possible to add whatever information/annotation you have or calculate directly to the `Spectra` object (as peaks or spectra variables) *before* plotting and then have a `labels()` function that can extract that information (or that even calculates it on the fly)? that was also the initial idea of accepting a function with parameter `labels`.

R/plotting-functions.R
guideflandre: This needs to be subject of discussion: 
1) `labels` could be of length 1 or equal to `x`, so if it is 1, repeat it `length(x)` times
2) only accept `labels` as a function (= `addFragments`) that returns a list of length equal `length(x)`, or accept `labels` as a list of length equal to `length(x)`. 

In the above, I used option 2) as I also used it for cases where variable modifications where used: which is also why there is an attribute to the elements of the returned list of `addFragments`. The attribute is an integer that corresponds to the spectrum that is called (if `plotSpectra(sp[1:4], labels = addFragments` is called, then `addFragments` returns a list of 4 elements. Each element can be matched to its corresponding spectrum thanks to the attribute)

R/plotting-functions.R
guideflandre: See the answer above. When variable modifications are called, the length of the returned list is longer than `length(x)`. It is thus needed to match the annotation to its corresponding spectrum.

R/plotting-functions.R
guideflandre: The essence of having `labels` return a list is twofold:

1) `addFragments` used to throw an error when multiple Spectra were called when asking for a modification. This was because `addFragments` only accepted Spectra objects of length 1. 
2) In my personal use case of variable modifications, returning a list was the best option I had (including said attribute).

Thus this fixes a current Error, and it sets potential for variable modification.

guideflandre:
> General comment: I don't know your specific use case, but would it not be possible to add whatever information/annotation you have or calculate directly to the `Spectra` object (as peaks or spectra variables) _before_ plotting and then have a `labels()` function that can extract that information (or that even calculates it on the fly)? that was also the initial idea of accepting a function with parameter `labels`.

I believe we could if the number of annotations don't exceed `length(x)`. In my use case, with variable modifications, I would have multiple annotations/spectrum which is not ideal.

guideflandre:
A discussion with Laurent led to a potential different idea: the creation of a package specifically for plotting spectra. I understand my use cases might not fit in well within the `plotSpectra` environment, and it might need its separate function/package. I believe this should be an option/discussed too.

lgatto:
@jorainer - we need to move on with this issue to be able to merge the new PSMatch functionality. Do we need another discussion.

R/plotting-functions.R
```diff
@@ -134,21 +136,22 @@
...
 #'     c("Frodo", "Bilbo", "Peregrin", "Samwise"))
+#'     
 #' ## Plot each peak in a different color
-#' plotSpectra(sp, labels = function(z) unlist(z$label),
+#' plotSpectra(sp, labels = sp$label,
 #'     col = list(1:5, 1:4))
 #'
 #' ## Plot a single spectrum specifying the label.
-#' plotSpectra(sp[2], labels = c("A", "B", "C", "D"))
+#' plotSpectra(sp[2], labels = list(c("Gimli", "Legolas", "Aragorn", "Gandalf")))
```
sgibb, line 154: Are these names better than just letters?

R/plotting-functions.R
```diff
@@ -217,10 +222,18 @@ plotSpectra <- function(x, xlab = "m/z", ylab = "intensity", type = "h",
         main <- rep(main[1], nsp)
     if (nsp > 1)
         par(mfrow = n2mfrow(nsp, asp = asp))
+    if (length(labels)) {
+        if (is.function(labels)) {
+            labels <- labels(x)
+        }
+        if (length(labels) != length(x))
+            stop("Please provide a list of annotations of length equal to 'x'.")
+    } else {labels <- NULL}
```
sgibb, line 231: Do we need this? Isn't the `if (length(labels))` statement already enough?

Maybe, we should change the default to `label = list()` instead of the current `label = character()`.

R/plotting-functions.R
```diff
@@ -259,9 +272,16 @@ plotSpectraOverlay <- function(x, xlab = "m/z", ylab = "intensity",
     if (frame.plot)
         box(...)
     title(main = main, xlab = xlab, ylab = ylab, ...)
+    if (length(labels)) {
+        if (is.function(labels)) {
+            labels <- labels(x)
+        }
+        if (length(labels) != length(x))
+            stop("Please provide a list of annotations of length equal to 'x'.")
+    } else {labels <- NULL}
```
sgibb, line 281: Do we need this? Isn't the `if (length(labels))` statement already enough?

Maybe, we should change the default to `label = list()` instead of the current `label = character()`.

R/plotting-functions.R
```diff
@@ -319,7 +345,7 @@ setMethod(
             xlim[1L] <- xlim[1L] - wdths
             xlim[2L] <- xlim[2L] + wdths
             plot.window(xlim = xlim, ylim = ylim, ...)
-        }
+        } else {labels <- NULL}
```
sgibb, line 348: Do we need this? Isn't the `if (length(labels))` statement already enough?

Maybe, we should change the default to `label = list()` instead of the current `label = character()`.

R/plotting-functions.R
```diff
@@ -217,10 +222,18 @@ plotSpectra <- function(x, xlab = "m/z", ylab = "intensity", type = "h",
...
+        }
+        if (length(labels) != length(x))
+            stop("Please provide a list of annotations of length equal to 'x'.")
+    } else {labels <- NULL}
+    
     for (i in seq_len(nsp))
         .plot_single_spectrum(x[i], xlab = xlab, ylab = ylab, type = type,
                               xlim = xlim, ylim = ylim, main = main[i],
-                              col = col[[i]], labels = labels,
+                              col = col[[i]], labels = labels[[i]],
```
sgibb, line 236: Maybe the default for `label` in `.plot_single_spectrum` should be changed as well?
(and instead of testing label in each of the other function it should be done once in `.plot_single_spectrum`).

sgibb (changes requested):
I am not aware of possible side effects of changing `labels = character()` to `labels = list()`. Did you try to rebuild the vignettes and reverse dependencies?

R/plotting-functions.R
guideflandre: Not at all, I saw them in another example and reused them. I'll change them again.

R/plotting-functions.R
guideflandre: Yes we should !

jorainer (approved):
I'm OK merging after Sebastian's comments have been addressed.

This might now however break code in other packages too - could you, for backward compatibility, maybe you could add a check to the functions

```r
if (is.character(labels))
    labels <- as.list(labels)
```

still, in the documentation etc I would not mention the support for `character`, this is purely for existing code in other packages or our analysis workflows that would now break, because we're passing a `character`.

R/plotting-functions.R
guideflandre: It had to come before because I needed to match the spectrum and the labels before `.plot_single_spectrum` is called. The default `labels` in `.plot_single_spectrum` thus still reamains a character.

guideflandre:
> This might now however break code in other packages too - could you, for backward compatibility, maybe you could add a check to the functions
> 
> ```r
> if (is.character(labels))
>     labels <- as.list(labels)
> ```
I changed into this: 
```r
if (is.character(labels))
    labels <- list(labels)
```
Since otherwise it would create a list of length equal to the number of peaks in the spectrum.

R/plotting-functions.R
guideflandre: Actually we do need it. If labels is an empty list, when calling `.plot_single_spectrum`, an error occurs: the subscript is out of bounds (`labels[[i]]` of an empty list). If labels is NULL, this error does not occur

guideflandre:
There was also an error with the old par not being returned after plotting multiple plots. I made sure plotSpectra now returns it

jorainer:
what puzzles me a bit is that the GHA are not running for this PR. Would be nice to actually see if all unit tests run smoothly

guideflandre:
> what puzzles me a bit is that the GHA are not running for this PR. Would be nice to actually see if all unit tests run smoothly

I ran the tests `test_plotting-functions.R` and they all pass when I remove the `old_par <- par(no.readonly = TRUE)` & `on.exit(par(old_par))`. 
I still believe this should be included. When I included them and compared the plots from the unit test, there are no visual changes whatsoever, only the par is different. When retaining the old par, it removes the annoyance of having a single plot be plotted in a [1,2] matrix after plotting 2 plots with plotSpectra. I could make a separate PR for this, but really it just removes this annoyance.

lgatto:
> Did you try to rebuild the vignettes and reverse dependencies?

I am trying to use the `revdepcheck` package, but so far without success - the supposed errors don't make any sense. Testing a couple of dependencies by hand works.

lgatto:
> I am trying to use the revdepcheck package, but so far without success - the supposed errors don't make any sense. 

It was apparently a timeout issue. Thinks are looking better now.

lgatto:
@sgibb - I confirm that the package and all revdeps check correctly.

lgatto (approved):

Issue #350, opened 2025-02-11
compareSpectra to report also the number of matching peaks


jorainer:
`compareSpectra()` performs the spectra similarity calculations in a two-step approach: first it matches (maps) peaks between the compared spectra and then it calculates the similarity of these. As a result, `compareSpectra()` returns a `numeric` `matrix` with the similarity score for all pairs of spectra. It might however also be interesting to report **in addition** the number of matched peaks between the compared spectra. The result could then be a `numeric` `array` with 3 dimensions, so that `sim[ , , 1L]` would be the similarity scores and `sim[, , 2L]` the number of common peaks between the matched spectra.

jorainer:
And the equivalent (updated) function to calculate similarities (modified from [.compare_spectra_chunk()](https://github.com/rformassspectrometry/Spectra/blob/main/R/Spectra-functions.R#L296-L355)) could be:

```r
.compare_spectra_chunk_npeaks <- function(x, y = NULL, MAPFUN = joinPeaks,
                                          tolerance = 0, ppm = 20,
                                          FUN = ndotproduct,
                                          chunkSize = 10000, ...,
                                          BPPARAM = bpparam(),
                                          peakCount = FALSE) {
    nx <- length(x)
    ny <- length(y)
    if (peakCount) {
        compare_fun <- .peaks_compare_npeaks
        dn <- list(spectraNames(x), spectraNames(y),
                   c("score", "peak_count"))
        dims <- c(nx, ny, 2L)
        z_chunk <- 1:2
    } else {
        compare_fun <- Spectra:::.peaks_compare
        dn <- list(spectraNames(x), spectraNames(y))
        dims <- c(nx, ny, 1L)
        z_chunk <- 1L
    }
    bppx <- backendBpparam(x@backend, BPPARAM)
    if (is(bppx, "SerialParam"))
        fx <- NULL
    else fx <- backendParallelFactor(x@backend)
    pqvx <- Spectra:::.processingQueueVariables(x)
    bppy <- backendBpparam(y@backend, BPPARAM)
    if (is(bppy, "SerialParam"))
        fy <- NULL
    else fy <- backendParallelFactor(y@backend)
    pqvy <- Spectra:::.processingQueueVariables(y)
    if (nx <= chunkSize && ny <= chunkSize) {
        mat <- compare_fun(
            Spectra:::.peaksapply(x, f = fx, spectraVariables = pqvx, BPPARAM = bppx),
[...]

PR #351, opened 2025-02-19
Add fillCoreSpectraVariables utility function and report number of matching peaks with compareSpectra

jorainer:
This PR adds two things:
- Add the `fillCoreSpectraVariables()` function that fills a spectra data frame with eventually missing core spectra variables (with the correct data type). This function will be particularly helpful for newly implemented `MsBackend` classes.
- Add parameter `matchedPeaksCount` parameter to `compareSpectra()` to allow reporting of the number of matching peaks between compared spectra (see issue #350 for details).


vignettes/Spectra.Rmd
```diff
@@ -1016,6 +1016,22 @@ being the two `Spectra` objects to compare. As a result a *n x m* matrix will be
...
+
+```{r}
+sim <- compareSpectra(sps, sps, ppm = 40, matchedPeaksCount = TRUE)
+
+#' The similarity scores:
+sim[, , 1L]
+
+#' The number of matching peaks:
+sim[, , 2L]
+```
```
philouail, line 1033: Maybe it's unnecessary, but maybe you could add a small example of filtering based on the number of matching peaks.

philouail (approved):
This looks really good, I'm all good with the code and documentation. 

I just added a small suggestion for the vignette.

vignettes/Spectra.Rmd
jorainer: hm, yes, thanks. I will add that later.

PR #354, opened 2025-03-04
feat: add parameter return.type to peaksData

jorainer:
This PR adds a new parameter `return.type` to `peaksData()`. With `return.type = "list"` the data would not be converted first to a `SimpleList` which has performance advantages (avoids copying of the data, but also operations such as `split()` etc are faster on a `list` than on a `SimpleList`). Thus, we might set `return.type = "list"` in other packages such as *xcms*.

### DIFF
--- a/R/Spectra.R
+++ b/R/Spectra.R
@@ -639,6 +639,10 @@ setReplaceMethod("dataStorageBasePath", "Spectra", function(object, value) {
 #'
 #' @param object A `Spectra` object.
 #'
+#' @param return.type For `peaksData()`: `character(1)` allowing to specify if
+#'     the results should be returned as a `SimpleList` or as a `list`.
+#'     Defaults to `return.type = "SimpleList"`.
+#'
 #' @param spectraVars `character()` indicating what spectra variables to add to
 #'     the `DataFrame`. Default is `spectraVariables(object)`, i.e. all
 #'     available variables.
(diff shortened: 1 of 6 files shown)


R/Spectra.R
```diff
@@ -1203,10 +1207,19 @@ setMethod("mz", "Spectra", function(object, f = processingChunkFactor(object),
...
     "peaksData", "Spectra",
     function(object, columns = c("mz", "intensity"),
-             f = processingChunkFactor(object), ..., BPPARAM = bpparam()) {
+             f = processingChunkFactor(object),
+             return.type = c("SimpleList", "list"), ..., BPPARAM = bpparam()) {
+        return.type <- match.arg(return.type)
         if (length(object@processingQueue) || length(f))
-            SimpleList(.peaksapply(object, columns = columns, f = f))
-        else SimpleList(peaksData(object@backend, columns = columns))
+            switch(return.type,
```
philouail, line 1214: dumb dev question, I though `switch()` was mostly used when you have strictly >2 option. but it's also more efficient than `if else` if we have 2 possible options ?

philouail (approved):
Completely approve this. Thanks for implementing.

R/Spectra.R
jorainer: I'm not a fan of `if` `else if` - that's why I used `switch` here :)

PR #359, opened 2025-07-02
Add precursorPurity function

Ahlammentag:
Add the precursorPurity function


jorainer:
Thanks for the review @philouail ! I will then - after your PR #362 - merge and fix the unit tests/Rd files.

jorainer:
Thanks @lgatto - no, it's not directly related to *msPurity* - but in fact the concept/calculation seems to be the same. I will add that to the documentation. Note that *msPurity* is unfortunately still using *MSnbase*, not *Spectra*.

guideflandre:
Is there a reason why `precScanNum()` was not called ? It allows to remove the vapply and the need to have the `Spectra` object sorted by rt. I had written it along the lines of :

```r
.precursorPurity2 <- function(x, tolerance = 0.3, ppm = 0,
                             useIsolationWindow = FALSE) {

    ## Get pairs of MS1 and MS2 spectra
    ms2_idx <- which(msLevel(x) == 2L)
    scans <- scanIndex(x) ## The scan indices of all spectra
    prec_scans <- precScanNum(x[ms2_idx]) ## Corresponding precursor scans
    ms1_idx <- match(prec_scans, scans)
...
}
```

instead of 

```r
.precursorPurity <- function(x, tolerance = 0.3, ppm = 0,
                             useIsolationWindow = FALSE) {
    if (is.unsorted(rtime(x)))
        stop("Spectra with data origin ", dataOrigin(x[1L]),
             " are not increasingly sorted by retention time.")
    ## Get pairs of MS1 and MS2 spectra
    ms2_idx <- which(msLevel(x) == 2L)
    ms1_all <- which(msLevel(x) == 1L)
    ms1_idx <- vapply(ms2_idx, function(i) max(ms1_all[ms1_all < i]), NA_real_)
    ms2_idx <- ms2_idx[is.finite(ms1_idx)]
    ms1_idx <- ms1_idx[is.finite(ms1_idx)]
```
I quickly benchmarked it and my implementation was indeed faster for smaller and bigger datasets.
I noticed though, that the dataset used in the example doesn't have any `precScanNum()` in which case the function would not work. I don't know if this is common as so far I've always had Spectra objects with it.

Ahlammentag:
The reason we didn't use precScanNum() is that it’s not available across all Spectra backends and data formats (in fact, it’s not available in the data I am working with). I think your version is faster in cases where you can guarantee that precScanNum() exists.

lgatto:
Two suggestions:
- What about checking if `precScanNum` is available and use it when possible, and otherwise, revert back to the current implementation?
- Or, possibly better, compute the `precScanNum` ourselves when missing (if possible), and then always use it?

Ping @jorainer

PR #381, opened 2026-03-16
feat: add rbindlistWithRownames function

jorainer:
- Add `rbindlistWithRownames()` function that uses `data.table::rbindlist()` for row binding `data.frame`s and conserves eventually present row names.
- Fix of unit tests based on latest updates in *MsCoreUtils*.


R/functions-util.R
```diff
@@ -140,3 +140,63 @@ sanitize_file_name <- function(x) {
...
+#'
+#' @param idcol Creates a column in the result showing which list item those
+#'     rows cam from. See [data.table::rbindlist()] for details.
+#'
+#' @param ignore.attr `logical(1)`, default `FALSE`. When `TRUE`, allows
+#'     binding columns with different attributes (e.g. class).
+#'
+#' @note
+#'
+#' No sanitizing of row names is performed, i.e., after merging duplicated
```
philouail: mm actually a data.frame cannot have dulplicated row.names no ?

tests/testthat/test_functions-util.R
```diff
@@ -79,3 +79,26 @@ test_that(".long_spectra_data3 works", {
...
+    rownames(b) <- paste0("b_", seq_len(nrow(b)))
+    expect_warning(res <- rbindlistWithRownames(list(a, b)), "not available")
+    expect_equal(nrow(res), nrow(a) + nrow(b))
+    expect_equal(rownames(res), as.character(seq_len(nrow(res))))
+
+    ## both have rownames
+    rownames(a) <- seq_len(nrow(a))
+    res <- rbindlistWithRownames(list(a, b))
+    expect_equal(nrow(res), nrow(a) + nrow(b))
+    expect_equal(rownames(res), c(rownames(a), rownames(b)))
```
philouail, line 102: I would do a test with duplicated rownames ?

philouail (approved):
Good for me, just one ocmmnet, i might be misunderstanding something.

R/functions-util.R
jorainer: good question. need to check that.

jorainer: ha, that's the beauty. by convention `row.names<-` does not support passing duplicated or NA row names. I used `attr(x, "row.names") <- ` which apparently did not check if there are duplicated names... I might need to change that :)

tests/testthat/test_functions-util.R
jorainer: yep! will do!

R/functions-util.R
jorainer: good point @philouail ! I changed the code now and drop rownames (with warning) if there are duplicated ones.

philouail (approved):
Perfect ! thanks for check the rownames things

PR #382, opened 2026-03-31
add method of "group_mz_int"

MuyaoXi9271:


jorainer:
@MuyaoXi9271 , can you please use 4 whitespaces (not 2)? otherwise also lines that you did not edit or change look as modified. 

this setting should actually be defined in the *.editorconfig* file - and ideally, your editor/IDE should read this file and set the configuration accordingly.

jorainer:
Dear @MuyaoXi9271 , can you please also describe what the goal/intention of your updates are? Maybe also have a look at the `reduceSpectra()` function which would report for groups of peaks with similar m/z the one with the highest intensity (for each spectrum)?

MuyaoXi9271:
> Dear @MuyaoXi9271 , can you please also describe what the goal/intention of your updates are? Maybe also have a look at the `reduceSpectra()` function which would report for groups of peaks with similar m/z the one with the highest intensity (for each spectrum)?

Dear @jorainer, 

Thanks for your suggestion. 

The DDA data acquired from the Bruker instrument from my lab is that the same precursor has several spectra at each collision energy (in total 5 ), so I managed to get a consensus spectrum for each precursor at each collision energy. When I use the existing group function within Spectra::combineSpectra like 
 `Spectra::combineSpectra(ex_spectra_filt, 
                                           FUN = combinePeaksData, 
                                           tolerance = 0.005,
                                           group_method = "group",
                                           peaks = intersect,
                                           minProp = 0.7,
                                           intensityFun = max, 
                                           mzFun = median,
                                           f = ex_spectra_filt$collisionEnergy)`_
I noticed that the difference in mass between smallest and largest within the same mass group is larger than the required threshold, so I include the condition of intensity to restrict how to group the masses, then to get the consensus spectrum for each precursor at each collision energy.

Best,
Muyao

jorainer:
OK, so the `group_mz_int()` is then doing a different (less *greedy*) grouping than `group()` that actually ensures the grouped peaks are **all** within `tolerance` and `ppm`? but this function is not included in this PR, right?

an additional question:

- I guess the problem is that the reported (consensus) m/z might be off when you report the median m/z of the peak group - but with the intensity weighted mean it should be somewhat similar to your result? because then the m/z closer to the max intensity peak will be reported.

MuyaoXi9271:
> OK, so the `group_mz_int()` is then doing a different (less _greedy_) grouping than `group()` that actually ensures the grouped peaks are **all** within `tolerance` and `ppm`? but this function is not included in this PR, right?
> 
> an additional question:
> 
> * I guess the problem is that the reported (consensus) m/z might be off when you report the median m/z of the peak group - but with the intensity weighted mean it should be somewhat similar to your result? because then the m/z closer to the max intensity peak will be reported.


I would say a little modification on `group()` to make sure the grouped peaks are **all** within `tolerance` and `ppm`. The function `group` is from the package **MsCoreUtils**, so I added the function `group_mz_int` to that package as well.

The idea of the function `group_mz_int` is to order the intensity of all masses from several spectra for the same precursor at the same collision energy, decreasingly,  then the mass with the highest intensity was picked up, and then other similar masses are found within a threshold. This step was repeated until picking the last mass group. In theory, the m/z closer to the max intensity from each mass group should be reported .

MuyaoXi9271:
Hi @jorainer , should I open a "pull request" under the package of MsCoreUtils?

jorainer:
OK, thanks for explaining. I think I understood it. And yes @MuyaoXi9271 , please make a PR to *MsCoreUtils* with your `group_mz_int()` function. Ideally, the function should then have the same result type than `group` (i.e. an integer with the assignment of the values to each group).

Then we can think how to also modify the existing `combinePeaks()` function to allow using different *grouping* functions (ideally without having to hard-code it using `if/else` or `switch` statements).

R/peak-list-functions.R
```diff
@@ -198,6 +198,7 @@ NULL
 combinePeaksData <- function(x, intensityFun = base::mean,
                              mzFun = base::mean, weighted = FALSE,
                              tolerance = 0, ppm = 0, timeDomain = FALSE,
+                             group_method = c("group", "group_mz_int"),
```
jorainer, line 201: Please use `groupMethod` instead of `group_method` to be consistent with the other parameter names. Also, please add the `groupMethod` parameter as last parameter (i.e., **after** `minProp`) to not cause problems if any user uses positional argument passing instead of named parameters.

R/peak-list-functions.R
```diff
@@ -198,6 +198,7 @@ NULL
...
                              peaks = c("union", "intersect"), main = integer(),
                              minProp = 0.5, ...) {
     peaks <- match.arg(peaks)
```
jorainer, line 204: you should also do a `groupMethod <- match.arg(groupMethod)` here so that the input parameter values are checked and a single one is chosen (by default the first).

R/peak-list-functions.R
```diff
@@ -225,8 +242,41 @@ combinePeaksData <- function(x, intensityFun = base::mean,
         mzs <- mzs[keep]
         ints <- ints[keep]
     }
+    
+    #New lines start-------- 
+    mzs <- mzs[mz_groups != -1]
```
jorainer, line 247: these lines (`mzs <- mzs[mz_groups != -1]` to `mz_groups <- mz_groups[mz_groups != -1]`) should go above inside the `else if (group_method == "group_mz_int")`. this is not needed for the `"group"` method and will slow the function down.

R/peak-list-functions.R
```diff
@@ -225,8 +242,41 @@ combinePeaksData <- function(x, intensityFun = base::mean,
...
+    ints <- ints[mz_groups != -1]
+    mz_groups <- mz_groups[mz_groups != -1]
+    #New lines end--------
     mzs <- split(mzs, mz_groups)
     ints <- split(ints, mz_groups)
+    
+    
+    # --- New lines start for keeping the mz within the tolerance ----------------------
```
jorainer, line 255: why is this whole code (line 255 until 276) needed? this is what the `group_mz_int()` should actually do, no?

R/peak-list-functions.R
```diff
@@ -238,12 +288,22 @@ combinePeaksData <- function(x, intensityFun = base::mean,
     if (weighted) {
         wm <- stats::weighted.mean
         mzs <- mapply(mzs, ints, FUN = function(mz, w)
-                      wm(mz, w + 1, na.rm = TRUE, USE.NAMES = FALSE))
+            wm(mz, w + 1, na.rm = TRUE, USE.NAMES = FALSE))
     } else
         mzs <- vapply1d(mzs, FUN  = mzFun)
+    
+    #New lines start ---------------------------------------------
```
jorainer, line 295: why are these changes here needed? `mzs` and `ints` where already ordered by m/z before, so why should they be re-ordered?

Update.txt
```diff
@@ -0,0 +1,3 @@
+Optimization of function of "peak-list-function.R" from lin 231 to line 252
```
jorainer, line 1: please remove the Update.txt file.

jorainer (changes requested):
Thanks for the PR. There are some changes and open questions that need to be addressed.

PR #383, opened 2026-04-02
feat: add shiftPeaks function

jorainer:
- Add function `shiftPeaks()` that allows to shift mass peaks in m/z dimension by an offset, which can be a single number or value(s) from a spectra variable.


R/Spectra.R
```diff
@@ -3127,6 +3145,10 @@ setMethod("filterValues", "Spectra",
 #'
 #' @param object A `Spectra` object.
 #'
+#' @param offset For `shiftPeaks()`: `numeric(1)` offset or `character(1)` with
+#'     the name of a spectra variable containing a (per spectrum) offset value
```
philouail: an* ?

R/Spectra.R
```diff
@@ -3241,6 +3263,19 @@ setMethod("filterValues", "Spectra",
...
+#' ## The `shiftPeaks()` function allows to shift peaks in each spectrum by
+#' ## an offset value in m/z dimension. As an example we below substract the
+#' ## precursor m/z value from the peaks' m/z of each MS2 spectrum to create
+#' ## spectra for a *neutral loss* comparison
+#'
+#' ## Add the negative precursor m/z as a new spectra variable
+#' sps_mod$precursor_offset <- -sps_mod$precursorMz
+#' nl_ms2 <- sps_mod |>
+#'     filterMsLevel(2L) |>
+#'     shiftPeaks(offset = "precursor_offset")
```
philouail, line 3275: probably dumb proposition but why not have a method and if a numeric() is provided then it is used to offset directly and if it is a character(1) then the method use the spectraData. ?or it is just too complicated to store for the lazy queueing ?

philouail (approved):
Looks good to me ! i found one typo and add one question :)

R/Spectra.R
jorainer: Hm. I see the point. maybe not having a method that dispatches on `offset` - the `if`/`else` isn't that much of a problem here, since there are only these two choices. I would make a `method` if it would need to dispatch for more cases. but indeed, why not having two different *peaks* functions, one that works with a numeric offset, and one that uses the info from the `spectraData`. this would avoid calling the `if` in the *peaks* function for each individual spectrum and should speed up the processing.

I will implement that and push.

PR #389, opened 2026-09-09
feat: ggplot2/ggiraph version of Spectra plot

gabrieletome:


lgatto:
Before looking into the details of the PR, could you motivate this addition? What does using ggplot bring compared to the existing base plotting functionality ? Could we envision creating a dedicated visualisation package for Spectra (this was already discussed in the past).

gabrieletome:
> Before looking into the details of the PR, could you motivate this addition? What does using ggplot bring compared to the existing base plotting functionality ? Could we envision creating a dedicated visualisation package for Spectra (this was already discussed in the past).

The ggplot functions are simply the ggplot version of the current plot, but these upgrades simplify obtaining interactive plots with ggiraph. I prefer ggiraph over plotly because of its tighter integration with ggplot2. While we can discuss moving everything into SpectraVis, I would prefer to keep the core plotting functions here and reserve SpectraVis strictly for the Shiny wrapper.

lgatto:
Does that mean that maintainers now need to maintain two versions, base and ggplot2, of functions that do the same? Will we have to guarantee that both are equivalent? What if one gets a feature that can't be (easily) implemented in the other?

lgatto:
My a priori opinion is that we should have a single set of default/simple plotting functions in Spectra (base or ggplot2, open for discussion), and offer alternatives in (an)other package(s), to facilitate maintenance and open up new development avenues.  I'm open to discussions, of course.

jorainer:
I have no strong opinion on where to add the functionality. I see the point that having duplicated plotting (both ggplot and base) doing the exact same within the same package does not make much sense. I could imagine that ggplot-based plots might be easier to customize. then let's maybe for now move the ggplot versions to *SpectraVis*. Maybe then also a vignette in *SpectraVis* comparing the two plotting approaches might be helpful.

lgatto:
I would also be open to keeping the gglot2 functions and moving (or deprecating) the base ones.

lgatto:
Maybe, to be able to make a more objective decision, it would be useful to have a short comparison of the two, and hint on possible future developments.

gabrieletome:
I'll compare the methods so we can make the best decision.

gabrieletome:
In this repository [gabrieletome/Spectra_test](https://github.com/gabrieletome/Spectra_test), I compared the functions for generating figures in the `Spectra` package. The comparison `plot` vs `ggplot2` focuses on two main points:

- *Installation footprint*: the `ggplot2` method requires more dependencies
  (134 vs 118), but the installation time is fairly similar.

- *Rendering of the figures*: all tests are described in the notebook. From
  these tests, I can summarize:
  - `plot` is faster in most cases; it is only slower than `ggplot2` a few
    times, when plotting large figures using `MsBackendMzR`.
  - `ggplot2` handles size limits better, despite a problem in visualization that 
    can be solved by changing the limits manually. `Base R plot` fails to render the test
    with an unusually high number of spectra (101 spectra plotted using
    `plotSpectra()`).
  - A major advantage of `ggplot2` is its straightforward compatibility with
    `ggiraph` for plot interactivity.
  - `ggplot2` is easier to customize outside the function.

To sum up, `base R plot` is simpler and faster, but it is more difficult to
customize the figures outside the function and to handle interactivity.
`ggplot2` requires more dependencies and is a bit slower, but it is simpler to
customize outside the function and to generate interactive plots.

If you want additional analysis, please let me know.

